PR #48, opened 2024-08-28
install-postfix.md: Remove noplaintext from sasl_security_options

nikolas:
The resulting `main.cnf` file in has `smtpd_sasl_security_options = noanonymous` rather than `smtpd_sasl_security_options = noanonymous,noplaintext`.

https://unix.stackexchange.com/a/765723/25861

This change seems to fix a postfix error I ran into when setting this up:
```
smtpd fatal: no SASL authentication mechanisms
```


bryceharrington:
Thanks, LGTM

PR #55, opened 2024-09-26
Add Profile-Guided Optimization page

sergiodj:
This adds a page explaining what Profile-Guided Optimization (PGO) is.  It also provides an example of an application being PGO'd.

I tried to hit a balance between explaining everything important in detail while not going into a deep dive on things that I don't consider worth it.  I realize that this is a complex topic, so I would appreciate a review.  Please let me know if there's anything that could be improved.

Thanks!


explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+On top of that, the current `autofdo` version in Ubuntu (`0.19-3build3`, at the time of this writing) is not recent enough to process the `perfdata` files we will generate.  There is a PPA with a newer version of `autofdo` package for Ubuntu Noble [here](https://launchpad.net/~sergiodj/+archive/ubuntu/autofdo).  If you are running another version of Ubuntu and want to install a newer version of `autofdo`, you will need to build the software manually (please refer to the [upstream repository](https://github.com/google/autofdo) for further instructions).
+
+## A simple PGO scenario: `openssl speed`
+
+PGO makes sense when your software is CPU-bound, i.e., when it performs CPU intensive work and is not waiting on I/O, for example.  OpenSSL has a built-in benchmark command called `openssl speed`, which tests the performance of its cryptographic algorithms.  This is excellent for PGO because there is practically no I/O involved, and we are just constrained by how fast the CPU can run.
```
cpaelzer, line 33: "not waiting" ->  "not waiting most of the time" or "not mostly waiting"?

A bit of I/O wait is fine.
And while I get it I wonder if we should mention/explain that PGO might still help, but you'd not see the effect due to it being hidden behind those waits?

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Running OpenSSL tests
+
+In order to measure the performance impact of PGO, it is important to perform several tests using an OpenSSL binary without PGO and then with it.
+
+The first thing we have to do is to do several runs of `openssl speed` *before* enabling PGO, so that we have a baseline to compare with.  As explained in the sections above, `autofdo` does not handle stripped binaries nor `dwz`-processed debug information, so we need to make
+After confirming that everything looks OK, we are ready to start the benchmark.  Let's run the command:
```
cpaelzer, line 40: This looks like it has eaten half of your sentence.
I assume you wanted to outline that you build it that way here?

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+On top of that, the current `autofdo` version in Ubuntu (`0.19-3build3`, at the time of this writing) is not recent enough to process the `perfdata` files we will generate.  There is a PPA with a newer version of `autofdo` package for Ubuntu Noble [here](https://launchpad.net/~sergiodj/+archive/ubuntu/autofdo).  If you are running another version of Ubuntu and want to install a newer version of `autofdo`, you will need to build the software manually (please refer to the [upstream repository](https://github.com/google/autofdo) for further instructions).
+
+## A simple PGO scenario: `openssl speed`
+
+PGO makes sense when your software is CPU-bound, i.e., when it performs CPU intensive work and is not waiting on I/O, for example.  OpenSSL has a built-in benchmark command called `openssl speed`, which tests the performance of its cryptographic algorithms.  This is excellent for PGO because there is practically no I/O involved, and we are just constrained by how fast the CPU can run.
+
+### Running OpenSSL tests
+
+In order to measure the performance impact of PGO, it is important to perform several tests using an OpenSSL binary without PGO and then with it.
```
cpaelzer, line 37: you are probably right to just call it "several tests".
But I wonder if somewhere, we should mention that this is to eliminate outliers, that people might want to consider disabling background load and all of it is about getting statistically worthwhile results

explanation/performance/perf-pgo.md
cpaelzer: you have "Building OpenSSL for profiling" so I assume you want to refer to that here

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+sha512           91713.74k   366632.96k   684349.30k  1060512.22k  1260359.00k  1273277.92k
+aes-256-cbc    1266018.36k  1437870.33k  1442743.14k  1449933.84k  1453336.84k  1453484.99k
+md5             140978.58k   367562.53k   714094.14k   955267.21k  1060969.81k  1068296.87k
+                  sign    verify    sign/s verify/s
+rsa 2048 bits 0.000173s 0.000012s   5776.8  86791.9
+```
+
+### Building OpenSSL for profiling
+
+Before we are able to profile OpenSSL, we need sure that we compile the software in a way that the generated program meets the requirements.  In our case, you can use the `file` command after building the binary to make sure that it has not been stripped:
```
cpaelzer, line 60: we know, but a random person might not immediately know how to "make sure that it has not been stripped:".

maybe you want to refer what you watch out for in the file output e.g. ` with debug_info, not stripped`

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+$ eu-readelf -S /usr/bin/openssl
+[...]
+Section Headers:
+[Nr] Name                 Type         Addr             Off      Size     ES Flags Lk Inf Al
+[...]
+[31] .debug_abbrev        PROGBITS     0000000000000000 001bcbf3 000110c4  0        0   0  1
+[...]
+```
+
+If the `Flags` field has `C` in it, this means that `dwz` was used.  You will then need to figure out how to disable its execution during the build.
```
cpaelzer, line 79: C as in "compressed" I guess?

cpaelzer: maybe worth a link to a manpage or such?

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+	-e 'cpu/event=0xc4,umask=0x0,name=ex_ret_brn_tkn/' \
+	--branch-any \
+	--all-cpus \
+	--output openssl-nonpgo.perfdata \
+	-- openssl speed -mr -seconds 60 -evp md5 sha512 rsa2048 aes-256-cbc
+```
+
+Note that the only thing that differs is the PMU event to be monitored.
+
+After the command has finished running, you should see a file named `openssl-nonpgo.perfdata` in your current directory.  This is the file we will now need to convert to `gcov`.
```
cpaelzer, line 107: You have experimented with long profiling, long profiling at lower sampling frequency to keep size under control and more. Some people might not have a nice 60 second test. Would it be worth to elude to the existence of frequency control and present some file size estimates?

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Converting `perfdata` to `gcov` with `autofdo`
+
+With `autofdo` installed, you can convert the `perfdata` generated in the last step to `gcov` by doing:
+
+```bash
+$ sudo chown ubuntu:ubuntu openssl-nonpgo.perfdata
```
cpaelzer, line 114: This is a consequence of perf needing sudo.
IMHO the chown and a so far missing explanation why we need it should thematically be in the section above.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+$ create_gcov \
+	--binary /usr/bin/openssl \
+	--gcov openssl-nonpgo.gcov \
+	--profile openssl-nonpgo.perfdata \
+	--gcov_version 2 \
+	-use_lbr false
+```
+
+`create_gcov` is verbose and will display several messages that may look like something is wrong; you should inspect its output and also make sure to check the `$?` shell variable to make sure that it has finished successfully.  You will then have a file called `openssl-nonpgo.gcov` in your current directory, which is the `gcov` file we will use to feed back to GCC when recompiling OpenSSL.  You can inspect the file by using the `dump_gcov` tool.
```
cpaelzer, line 123: "several messages that may look like something is wrong"

can we assist the reader with examples what is "looking like but not bad" and what is "really bad"?

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+After the software is built, you can install this new version and run the benchmark tests again to compare results.
+
+### Our results
+
+We obtained some interesting results while performing the tests.  Here are the highlights:
+
+* Overall, the performance gain/loss was minimal.  Most of the time it was less than 1%.
+
+* There were some huge outliers, though.  For example, `sha512` showed a gain of more than 13% in one of the runs, but also showed a loss of 11% in another run!  This is likely noise due to external factors, even though we tried to isolate our test environment as much as possible.  Still, it was surprising to see such discrepancies.
```
cpaelzer, line 147: the noise as an excuse is not really helpful, but we can make it helpful.
For example we can show "but by disabling this and that in background" and "by running it N times excluding the outliers" and ... we have found this to be ...
And then, as you did well in touching but not explaining all elude to more like "but this is about geting reliable statistical data out of measurements which shall not be the topic here in all its detail"

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+Also note that enabling PGO will make GCC automatically enable several optimization passes that are usually disable, even when using `-O2`.  This can lead to new warnings/errors in the code, which might require either the code to be adjusted or the use of extra compilation flags to suppress these warnings.
+
+After the software is built, you can install this new version and run the benchmark tests again to compare results.
+
+### Our results
+
+We obtained some interesting results while performing the tests.  Here are the highlights:
+
+* Overall, the performance gain/loss was minimal.  Most of the time it was less than 1%.
```
cpaelzer, line 145: That as the first message is a bit non-inspiring.

We might want to mention that this is due to these hot code paths already being some that are highly optimized (they are) and that you've found with more complex software (in your other tests much more than a hand optimized hot loop is active) you've seen usually 5%-15%

Without framing it this way of "we chosen this for simplicity of the test, but thereby accepting we won't see much - yet you can have much we are shooting the messaging and helpfulness in the foot".

Once the blog exists which is more reporting what we did vs showing how one could recreate the same you can even point to it from here.

cpaelzer (commented):
I like it a lot, yet still have many suggestions.

Give them a thought and I hope they will help you to make this even more polished and worthwhile for the reader.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
+(perf-pgo)=
+
+# Profile-Guided Optimization
+
+[Profile-guided optimization](https://en.wikipedia.org/wiki/Profile-guided_optimization) (PGO) or Feedback-driven optimization (FDO) are synonyms for the same concept.  The idea is that a binary is compiled with built-in instrumentation which, when executed, will generate a profile.  This profile can then be used as an input for the compiler in a subsequent build of the same binary, and it will serve as guide for further optimizations.
```
s-makin, line 5:
```suggestion
[Profile-guided optimization](https://en.wikipedia.org/wiki/Profile-guided_optimization) (PGO) or feedback-driven optimization (FDO) are synonyms for the same concept.  The idea is that a binary is compiled with built-in instrumentation which, when executed, will generate a profile.  This profile can be used as input for the compiler in a subsequent build of the same binary, and it will serve as a guide for further optimizations.
```

explanation/performance.rst
```diff
@@ -38,10 +38,12 @@ following sections.
 * Obtain the hierarchical map of key computing elements using :ref:`hwloc and lstopo <perf-tune-hwloc>`
 * Check and control CPU governors, power and frequency with :ref:`cpupower <perf-tune-cpupower>`
 * Dynamic, adaptive system tuning :ref:`with TuneD <perf-tune-tuned>`
+* Profile-Guided Optimization :ref:`PGO <perf-pgo>`
```
s-makin, line 41: I actually think this new page should be at the top of the list

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+Ideally, the profile should be generated by a representative workload of the program.  Moreover, the built-in instrumentation impacts the overall performance of the binary.  Both things can make it hard to do profiling of real world applications, because it's not always possible to simulate a representative workload or because of the performance penalty.
```
s-makin, line 7:
```suggestion
It can be hard to do profiling of real world applications. Ideally, the profile should be generated by a representative workload of the program, but it's not always possible to simulate a representative workload.  Moreover, the built-in instrumentation impacts the overall performance of the binary which introduces a performance penalty.
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+* Please note that, despite being a relatively popular technique, PGO is not always the best approach to optimize a program.  The profiling data generated by the workload will be extremely tied to it, which means that the optimized program might actually have worse performance when other types of workloads are executed.  There is not a one-size-fits-all solution for this problem, and sometimes the best approach might be to **not** use PGO after all.
+
+* If you planning to follow along, we recommend setting up a test environment for this experiment.  The ideal setup involves using a bare metal machine because it's the more direct way to collect the performance metrics.  If you would like to use a virtual machine (created using QEMU/libvirt, LXD, Multipass, etc.), it will likely only work on Intel-based processors due to how vPMU (Virtual Performance Monitoring Unit) works.
```
s-makin, line 17:
```suggestion
* If you plan to follow along, we recommend setting up a test environment for this experiment.  The ideal setup involves using a bare metal machine because it's the more direct way to collect the performance metrics.  If you would like to use a virtual machine (created using QEMU/libvirt, LXD, Multipass, etc.), it will likely only work on Intel-based processors due to how vPMU (Virtual Performance Monitoring Unit) works.
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+## `perf` and AutoFDO
+
+Using `perf` to monitor a process and obtain data about its runtime workload means that we will end up with data files in a specific binary format that we will call `perfdata`.  Unfortunately, GCC doesn't understand this file format; instead, it expects a profile file that is in a format called `gcov`.  To convert a `perfdata` file into a `gcov` one, we will need to use a software called [`autofdo`](https://github.com/google/autofdo).  This software expects the binary being profiled to obey certain constraints:
```
s-makin, line 21:
```suggestion
Using `perf` to monitor a process and obtain data about its runtime workload produces data files in a specific binary format that we will call `perfdata`.  Unfortunately, GCC doesn't understand this file format; instead, it expects a profile file in a format called `gcov`.  To convert a `perfdata` file into a `gcov` one, we need to use a software called [`autofdo`](https://github.com/google/autofdo).  This software expects the binary being profiled to obey certain constraints:
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+* The binary **cannot** be stripped of its debug symbols.  `autofdo` does not support separate debug information files (i.e., it can't work with Ubuntu's `.ddeb` packages), and virtually all Ubuntu packages run `strip` doing their build in order to generate the `.ddeb` packages.  If you intend to profile an Ubuntu package, please keep that in mind.
+
+* The debug information file(s) **cannot** be processed by `dwz`.  This is a tool whose purpose is to compress the debug information generated when building a binary, and again, virtually all Ubuntu packages use it.  For this reason, it is currently not possible to profile most of Ubuntu's packages without first rebuilding them to disable `dwz` from running.
```
s-makin, line 25:
```suggestion
* The debug information file(s) **cannot** be processed by `dwz`.  This tool's purpose is to compress the debug information generated when building a binary, and again, virtually all Ubuntu packages use it.  For this reason, it is currently not possible to profile most Ubuntu packages without first rebuilding them to disable `dwz` from running.
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+* We have to be mindful of the options we pass to `perf`, particularly when it comes to recording branch prediction events.  The options will likely vary depending on whether you are using an Intel or AMD processor, for example.
```
s-makin, line 27:
```suggestion
* We must be mindful of the options we pass to `perf`, particularly when it comes to recording branch prediction events.  The options will likely vary depending on whether you are using an Intel or AMD processor, for example.
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+On top of that, the current `autofdo` version in Ubuntu (`0.19-3build3`, at the time of this writing) is not recent enough to process the `perfdata` files we will generate.  There is a PPA with a newer version of `autofdo` package for Ubuntu Noble [here](https://launchpad.net/~sergiodj/+archive/ubuntu/autofdo).  If you are running another version of Ubuntu and want to install a newer version of `autofdo`, you will need to build the software manually (please refer to the [upstream repository](https://github.com/google/autofdo) for further instructions).
```
s-makin, line 29:
```suggestion
On top of that, the current `autofdo` version in Ubuntu (`0.19-3build3`, at the time of this writing) is not recent enough to process the `perfdata` files we will generate.  There is a PPA with a newer version of `autofdo` package [for Ubuntu Noble](https://launchpad.net/~sergiodj/+archive/ubuntu/autofdo).  If you are running another version of Ubuntu and want to install a newer version of `autofdo`, you will need to build the software manually (please refer to the [upstream repository](https://github.com/google/autofdo) for further instructions).
```

explanation/performance/perf-pgo.md
s-makin: Or at least something that explains how to disable it during builds

explanation/performance/perf-pgo.md
s-makin: Examples are always good :)

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,153 @@
...
+
+How you perform this `CFLAGS` adjustment depends on how you built the software in the first place, so we won't cover this part here.  The resulting `CFLAGS` variable should have the following GCC option, though:
+
+```
+-fauto-profile=/path/to/openssl-nonpgo.gcov
+```
+
+Make sure to adjust the path to the `openssl-nonpgo.gcov` file.
+
+Also note that enabling PGO will make GCC automatically enable several optimization passes that are usually disable, even when using `-O2`.  This can lead to new warnings/errors in the code, which might require either the code to be adjusted or the use of extra compilation flags to suppress these warnings.
```
s-makin, line 137:
```suggestion
Also note that enabling PGO will make GCC automatically enable several optimization passes that are usually disabled, even when using `-O2`.  This can lead to new warnings/errors in the code, which might require either the code to be adjusted or the use of extra compilation flags to suppress these warnings.
```

explanation/performance/perf-pgo.md
s-makin: If someone is following along, won't their results depend on their hardware? If you're presenting a blog with the results of your testing, I agree that those should go here as an example of what could be seen. Since during the article you're using we/our throughout, it's not necessarily clear that "our results" here actually refers to the results of your testing

explanation/performance/perf-pgo.md
s-makin: It could be helpful to suggest an order of magnitude number of tests that could be needed to provide a decent signal-to-noise ratio. I assume you'll be reporting the number of tests you ran in your blog post, but still I think it's worth reporting on what could be considered a statistically useful number of tests

s-makin (commented):
Really nice article! I agree with all the points cpaelzer has already raised. Added a few extra suggestions as well.

explanation/performance/perf-pgo.md
sergiodj: Agreed.  The new text reads:

```
PGO makes more sense when your software is CPU-bound, i.e., when it performs CPU intensive work and is not mostly waiting on I/O, for example.  Even if your software spends time waiting on I/O, using PGO might still be helpful; its effects would be less noticeable, though.
```

explanation/performance/perf-pgo.md
sergiodj: Argh, indeed.  Sorry about that.  Referred the next section and fixed the sentence.

explanation/performance/perf-pgo.md
sergiodj: ACK.  Done.

explanation/performance/perf-pgo.md
sergiodj: Yes, `C` as in "compressed".  Unfortunately there's no mention of these flags in the official documentation.  elfutils unfortunately leaves a lot to be desired when it comes to discoverability :-/.  I've added `(as in compressed)` after the mention to `C`, hopefully that's clearer already.

explanation/performance/perf-pgo.md
sergiodj: Good point.  I've expanded the sentence, let me know what you think.

explanation/performance.rst
sergiodj: ACK.  Moved to the top :-).

explanation/performance/perf-pgo.md
sergiodj: Good idea.  For the file size estimates, I couldn't find my previous runs (I guess they were on horsea...), so I had to dig my notes/memory here.  I wasn't able to find a concrete value, so I ended up putting "several gigabytes" (which is true).  Hopefully that's good enough.

explanation/performance/perf-pgo.md
sergiodj: I've rearranged some parts of the text to make this clearer.

explanation/performance/perf-pgo.md
sergiodj: Sure thing.  I put an example of what a successful invocation looks like.

explanation/performance/perf-pgo.md
sergiodj: @cpaelzer and @s-makin, thanks for the suggestions.

@cpaelzer, would you have some time to look deeper into this with me?  I'd like your input regarding the statistical side of things, especially when it comes to determining a good number of tests to be run.

Unfortunately I have already released the machine I was using for testing this, so if more tests are needed they will take some time.

explanation/performance/perf-pgo.md
sergiodj: @cpaelzer Good suggestion.  I'll add some text expanding on the reason why we've seen such a low performance gain.

@s-makin The results will depend on a bunch of things, including their hardware, yes.  I still have to prepare the blog post, but I'll certainly link it here once it's done.

sergiodj:
BTW, after reading all your comments I've been thinking whether using `openssl speed` as an example was indeed a good idea, given that it did not provide a concrete example of optimization generated by PGO.  WDYT?

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,252 @@
...
+
+It can be hard to do profiling of real world applications. Ideally, the profile should be generated by a representative workload of the program, but it's not always possible to simulate a representative workload.  Moreover, the built-in instrumentation impacts the overall performance of the binary which introduces a performance penalty.
+
+In order to address these problems, nowadays we use tools like `perf` to "observe" what the binary is doing externally (sampling it, by monitoring events using Linux kernel's PMU --- Performance Monitoring Unit), which makes the process more suitable to be used in production environments.  This technique works better than the regular built-in instrumentation, but it still has a few drawbacks that we will expand later.
```
s-makin, line 9:
```suggestion
In order to address these problems, nowadays we use tools like `perf` to "observe" what the binary is doing externally (sampling it, by monitoring events using Linux kernel's Performance Monitoring Unit -- PMU), which makes the process more suitable to be used in production environments.  This technique works better than the regular built-in instrumentation, but it still has a few drawbacks that we will expand later.
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Caveats
+
+* The purpose of this guide is to provide some basic information about what PGO is.  In order to do that, we will look at a simple example using OpenSSL (more specifically, the `openssl speed` command) and learn how to do basic profiling.  We will not go into a deep dive on how to build the project, and it is assumed that the reader is comfortable with compilation, compiler flags and using the command line.
```
s-makin, line 13:
```suggestion
* The purpose of this guide is to provide some basic information about what PGO is and how it works.  In order to do that, we will look at a simple example using OpenSSL (more specifically, the `openssl speed` command) and learn how to do basic profiling.  We will not go into a deep dive on how to build the project, and it is assumed that the reader is comfortable with compilation, compiler flags and using the command line.
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,252 @@
...
+
+* Please note that, despite being a relatively popular technique, PGO is not always the best approach to optimize a program.  The profiling data generated by the workload will be extremely tied to it, which means that the optimized program might actually have worse performance when other types of workloads are executed.  There is not a one-size-fits-all solution for this problem, and sometimes the best approach might be to **not** use PGO after all.
```
s-makin, line 15:
```suggestion
* Despite being a relatively popular technique, PGO is not always the best approach to optimize a program.  The profiling data generated by the workload will be extremely tied to it, which means that the optimized program might actually have worse performance when other types of workloads are executed.  There is not a one-size-fits-all solution for this problem, and sometimes the best approach might be to **not** use PGO after all.
```

s-makin: There's a move away from using "please" in documentation.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,252 @@
...
+
+* If you plan to follow along, we recommend setting up a test environment for this experiment.  The ideal setup involves using a bare metal machine because it's the more direct way to collect the performance metrics.  If you would like to use a virtual machine (created using QEMU/libvirt, LXD, Multipass, etc.), it will likely only work on Intel-based processors due to how vPMU (Virtual Performance Monitoring Unit) works.
```
s-makin, line 17:
```suggestion
* If you plan to follow along, we recommend setting up a test environment for this experiment.  The ideal setup involves using a bare metal machine because it's the more direct way to collect the performance metrics.  If you would like to use a virtual machine (created using QEMU/libvirt, LXD, Multipass, etc.), it will likely only work on Intel-based processors due to how Virtual Performance Monitoring Unit (vPMU) works.
```

s-makin: It's becoming preferred to put the acronym in brackets after the full name of whatever the thing is - putting the acronym first implies that the reader ought to already know what the acronym stands for

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,252 @@
...
+
+## `perf` and AutoFDO
+
+Using `perf` to monitor a process and obtain data about its runtime workload produces data files in a specific binary format that we will call `perfdata`.  Unfortunately, GCC doesn't understand this file format; instead, it expects a profile file in a format called `gcov`.  To convert a `perfdata` file into a `gcov` one, we need to use a software called [`autofdo`](https://github.com/google/autofdo).  This software expects the binary being profiled to obey certain constraints:
```
s-makin, line 21: What's GCC?

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,252 @@
...
+
+* The binary **cannot** be stripped of its debug symbols.  `autofdo` does not support separate debug information files (i.e., it can't work with Ubuntu's `.ddeb` packages), and virtually all Ubuntu packages run `strip` doing their build in order to generate the `.ddeb` packages.  If you intend to profile an Ubuntu package, please keep that in mind.
```
s-makin, line 23:
```suggestion
* The binary **cannot** be stripped of its debug symbols.  `autofdo` does not support separate debug information files (i.e., it can't work with Ubuntu's `.ddeb` packages), and virtually all Ubuntu packages run `strip` during their build in order to generate the `.ddeb` packages.
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,252 @@
...
+
+On top of that, the current `autofdo` version in Ubuntu (`0.19-3build3`, at the time of this writing) is not recent enough to process the `perfdata` files we will generate.  There is a PPA with a newer version of `autofdo` package for Ubuntu Noble [here](https://launchpad.net/~sergiodj/+archive/ubuntu/autofdo).  If you are running another version of Ubuntu and want to install a newer version of `autofdo`, you will need to build the software manually (please refer to the [upstream repository](https://github.com/google/autofdo) for further instructions).
+
+## A simple PGO scenario: `openssl speed`
+
+PGO makes more sense when your software is CPU-bound, i.e., when it performs CPU intensive work and is not mostly waiting on I/O, for example.  Even if your software spends time waiting on I/O, using PGO might still be helpful; its effects would be less noticeable, though.
+
+OpenSSL has a built-in benchmark command called `openssl speed`, which tests the performance of its cryptographic algorithms.  This is excellent for PGO because there is practically no I/O involved, and we are just constrained by how fast the CPU can run.
```
s-makin, line 35:
```suggestion
OpenSSL has a built-in benchmark command called `openssl speed`, which tests the performance of its cryptographic algorithms.  This is excellent for PGO because there is practically no I/O involved, and we are only constrained by how fast the CPU can run.
```

s-makin: ("only" is less ambiguous than "just")

s-makin (commented):
Additional review on the parts that have been changed

explanation/performance/perf-pgo.md
cpaelzer: thanks for the discussion - go for:
1. positive tone
2. explain that non-improvement can happen even if doing right
3. hint at any general statistic trick to help like what sally and I mentioned
4. link/hint at blog post for an example of the same with improvements

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,252 @@
...
+$ eu-readelf -S /usr/bin/openssl
+[...]
+Section Headers:
+[Nr] Name                 Type         Addr             Off      Size     ES Flags Lk Inf Al
+[...]
+[31] .debug_abbrev        PROGBITS     0000000000000000 001bcbf3 000110c4  0        0   0  1
+[...]
+```
+
+If the `Flags` field has `C` (as in `compressed`) in it, this means that `dwz` was used.  You will then need to figure out how to disable its execution during the build.
```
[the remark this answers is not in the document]
sergiodj: It's as easy as not invoking `dwz` during the build.  It's just hard to write a generic way to do that without knowing the particularities of the build system being used.

explanation/performance/perf-pgo.md
sergiodj: The compiler.  A user reading this documentation will certainly know what it is, because it's an advanced topic.

explanation/performance/perf-pgo.md
sergiodj: Thanks @cpaelzer .  I've updated the last sections, please take a look when you can.

explanation/performance/perf-pgo.md
cpaelzer: ack to the explanation of @sergiodj , but since any question is a valid question maybe add a link to [GCC](https://gcc.gnu.org/) when we say GCC the first time?

explanation/performance/perf-pgo.md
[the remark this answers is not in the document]
cpaelzer: Maybe thereby lead in the second sentence with "Since the particularities how to skip that depend on the build system you are using [You will then need ...]"?

If you WANT to go deeper you might say - "For my build this was done via ... "

cpaelzer:
> BTW, after reading all your comments I've been thinking whether using `openssl speed` as an example was indeed a good idea, given that it did not provide a concrete example of optimization generated by PGO. WDYT?

As I said, yes it isn't the most catchy example. But many readers might do the same mistake by picking an already highly optimized case and then be sad to not see benefit. PGO has the biggest potential if there was not 25 years of optimization yet. Hence you can use your mistake to the benefit of others - explain that it can be normal to not see a benefit if e.g. the workload is not good for it (similar if things are entriely I/O bound).

And for motivation then Refer to your blog post which has examples where it really helped despite the case being quite complex, too complex for manual tuning for example.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,260 @@
...
+
+It can be hard to do profiling of real world applications. Ideally, the profile should be generated by a representative workload of the program, but it's not always possible to simulate a representative workload.  Moreover, the built-in instrumentation impacts the overall performance of the binary which introduces a performance penalty.
+
+In order to address these problems, nowadays we use tools like `perf` to "observe" what the binary is doing externally (sampling it, by monitoring events using Linux kernel's Performance Monitoring Unit -- PMU), which makes the process more suitable to be used in production environments.  This technique works better than the regular built-in instrumentation, but it still has a few drawbacks that we will expand later.
```
cpaelzer, line 9: "nowadays we use..." might age badly how about "we can use..."

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,260 @@
...
+
+On top of that, the current `autofdo` version in Ubuntu (`0.19-3build3`, at the time of this writing) is not recent enough to process the `perfdata` files we will generate.  There is a PPA with a newer version of `autofdo` package [for Ubuntu Noble](https://launchpad.net/~sergiodj/+archive/ubuntu/autofdo).  If you are running another version of Ubuntu and want to install a newer version of `autofdo`, you will need to build the software manually (please refer to the [upstream repository](https://github.com/google/autofdo) for further instructions).
+
+## A simple PGO scenario: `openssl speed`
+
+PGO makes more sense when your software is CPU-bound, i.e., when it performs CPU intensive work and is not mostly waiting on I/O, for example.  Even if your software spends time waiting on I/O, using PGO might still be helpful; its effects would be less noticeable, though.
+
+OpenSSL has a built-in benchmark command called `openssl speed`, which tests the performance of its cryptographic algorithms.  This is excellent for PGO because there is practically no I/O involved, and we are only constrained by how fast the CPU can run.
```
cpaelzer, line 35: Here you might hint at your later summary saying that it is a common mistake to use benchmarks and highly optimized code by saying something like "To be honest, benchmarks might already be highly optimized which again could cause issues, more about that later. After you got comfortable following our example the real benefit for you comes from using the same methods learned then against your own software stack".

That way we would nicely open and close that thought. And point to the example always being just that - an example. And the real win for the reader is to optimize what they have control about.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,260 @@
...
+
+```bash
+$ sudo chown ubuntu:ubuntu openssl-nonpgo.perfdata
+```
+
+Now, we need to convert the file to `gcov`.
+
+#### Alternative scenario: profiling the program using other approaches
+
+We do not always have well behaved programs that can be profiled like `openssl speed`.  Sometimes, the program might take a long time to finish running the workload, which ends up generating huge `perfdata` files that can prove very challenging for `autofdo` to process.  For example, we profiled QEMU internally and, when executing a workload that took around 3 hours to complete, the `perfdata` generated had a size of several gigabytes.
```
cpaelzer, line 130: When you say "qemu internally" you might also refer to the related blog post.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,260 @@
...
+$ create_gcov \
+	--binary /usr/bin/openssl \
+	--gcov openssl-nonpgo.gcov \
+	--profile openssl-nonpgo.perfdata \
+	--gcov_version 2 \
+	-use_lbr false
+```
+
+`create_gcov` is verbose and will display several messages that may look like something is wrong.  For example, the following output is actually from a successful run of the command:
+
```
cpaelzer, line 163: I know you want to show that even a successful run can throw these, but that is too much.
IMHO the first 6 lines, then ... and then the last 5 lines should be much better.
Otherwise this doesn't even fit on a full-hd screen.

So a) reduce lines please

But also reduce width as this need to scroll a lot left and right.
I think you could strip "third_party/perf_data_converter/src/" from the output without taking away anything of the important message you are telling. Some scrolling will remain, but people can see the important bit right away

cpaelzer: below you have used "[...] many more lines [...]" - the same works here to stay consistent

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,260 @@
...
+```
+-fauto-profile=/path/to/openssl-nonpgo.gcov
+```
+
+Make sure to adjust the path to the `openssl-nonpgo.gcov` file.
+
+Also note that enabling PGO will make GCC automatically enable several optimization passes that are usually disabled, even when using `-O2`.  This can lead to new warnings/errors in the code, which might require either the code to be adjusted or the use of extra compilation flags to suppress these warnings.
+
+After the software is built, you can install this new version and run the benchmark tests again to compare results.
```
cpaelzer, line 238: Do you think it makes sense to re-check if it was used via anything mentioned in https://stackoverflow.com/questions/12112338/get-the-compiler-options-from-a-compiled-executable ?
If you think not, feel free to ignore this is a low prio suggestion.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Our results
+
+At first sight it might appear that our results don't have much to show for themselves, but they are actually very interesting and allow us to discuss some more things about PGO.
+
+First, it is not a silver bullet.  It will not work for every scenario, and there might be cases when it is not justified to use it.  For example, if you are dealing with a very generic program that doesn't have a single representative workload.  PGO is great when applied to specialized cases, but may negatively impact the performance of programs that can be used in a variety of ways.
+
+It is also really important to be rigorous when doing the profiling of the program you want to optimize.  Here are a few tips that might help obtain more confident results:
+
+* Make sure to perform multiple runs of the same workload.  This helps to eliminate possible outliers as well as decrease the contribution of external factors (like memory and cache times).  In our example, we performed 5 consecutive runs without PGO and 5 consecutive runs with PGO.  Had we performed more runs, some of the outliers we see in the results would be even less noticeable.
```
cpaelzer, line 248: I like the new outro much more - thanks.

But you still say "... we see in the results" yet no more showing them as they are un-impressive. This has been about explaining the method for which this was great, so maybe avoid saying what we see in the results, but keep the rest here?

cpaelzer, line 248: Speaking of consecutive runs, that is true for measuring as well as for profiling - might mention that too.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,260 @@
...
+
+* Try to configure your system to have as little load as possible.  If you have other programs running in the background, they will certainly affect the final measurements in non-deterministic ways.
```
cpaelzer, line 250: Thanks for mentioning background load, not sure we should bring it up here but here is the catch. If we talk about rare spiky background like unattended upgrades that is all fine. But if in reality the load will run on a server dominated by a few other cache eating processes, then taking away the background load might make the program only better in the artificial environment. Due to that there is bad background load (spiky, non representative for most of the execution time, affecting measurement) and good background load (happening most of the execution time, representing the real environment this will be needed in production).

Up to you if you want to extend on that or not. It is ok without unless you want to add that.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,260 @@
...
+
+* Another important thing to consider is whether you are dealing with code that has already been highly optimized.  In fact, this is exactly the case with our example: OpenSSL is an ubiquitous cryptographic library that has received *a lot* of attention from several developers, and it is expected that cryptographic libraries are extremely optimized.  This is the major factor that explains why our results were not so great: the library is already pretty performant as it is!
```
cpaelzer, line 252: I think the reference to your "blog that shows actual benefit" needs to go here readers will memorize it here as  "not great" before coming to the "more info" a bit below. So I'd suggest to refer to it up here as well in the "our results" in addition to mentioning it below as further read.

cpaelzer (approved):
I've before replied suggestions and opinion on a few threads that have been open between you and Sally. Then I've done another full pass leaving a few hints.

But these are polishing, this is already awesome and hence you get my Approval even before considering my next barrage of suggestions. If you implement them, fine - if not for good reason this is still good to go and help the world that want to experiment with it.

explanation/performance/perf-pgo.md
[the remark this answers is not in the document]
sergiodj: Done.

explanation/performance/perf-pgo.md
sergiodj: Done.

explanation/performance/perf-pgo.md
sergiodj: Good catch.  Done.

explanation/performance/perf-pgo.md
sergiodj: +1.  I will publish this documentation first, but I'll amend it later with links to the blog post.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,260 @@
...
+
+Also note that enabling PGO will make GCC automatically enable several optimization passes that are usually disabled, even when using `-O2`.  This can lead to new warnings/errors in the code, which might require either the code to be adjusted or the use of extra compilation flags to suppress these warnings.
+
+After the software is built, you can install this new version and run the benchmark tests again to compare results.
+
+### Our results
+
+At first sight it might appear that our results don't have much to show for themselves, but they are actually very interesting and allow us to discuss some more things about PGO.
+
+First, it is not a silver bullet.  It will not work for every scenario, and there might be cases when it is not justified to use it.  For example, if you are dealing with a very generic program that doesn't have a single representative workload.  PGO is great when applied to specialized cases, but may negatively impact the performance of programs that can be used in a variety of ways.
```
s-makin, line 244:
```suggestion
First, it is not a "cure all".  It will not work for every scenario, and there might be cases when it is not justified to use it.  For example, if you are dealing with a very generic program that doesn't have a single representative workload.  PGO is great when applied to specialized cases, but may negatively impact the performance of programs that can be used in a variety of ways.
```

s-makin: "cure all" is more likely to carry your meaning than "silver bullet" across language divides -- tbh we could remove that part completely and just say "first, it will not work for..." to avoid the need for idioms, but it's not a blocker.

explanation/performance/perf-pgo.md
s-makin, line 248:
```suggestion
* Perform multiple runs of the same workload.  This helps to eliminate possible outliers and decrease the contribution of external factors (like memory and cache times).  In our example, we performed 5 consecutive runs without PGO and 5 consecutive runs with PGO.  Had we performed more runs, some of the outliers we see in the results would be even less noticeable.
```

explanation/performance/perf-pgo.md
s-makin, line 252:
```suggestion
* Consider whether you are dealing with code that has already been highly optimized.  In fact, this is exactly the case with our example: OpenSSL is a ubiquitous cryptographic library that has received *a lot* of attention from several developers, and it is expected that cryptographic libraries are extremely optimized.  This is the major factor that explains why our results were not so great: the library is already pretty performant as it is!
```

s-makin: I'm not sure it's fair to say that the results are "not so great" - I'd suggest framing it more as that you saw minimal performance gains or something like that.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,260 @@
...
+
+* Another important thing to consider is whether you are dealing with code that has already been highly optimized.  In fact, this is exactly the case with our example: OpenSSL is an ubiquitous cryptographic library that has received *a lot* of attention from several developers, and it is expected that cryptographic libraries are extremely optimized.  This is the major factor that explains why our results were not so great: the library is already pretty performant as it is!
+
+### Conclusion
+
+PGO is an interesting and complex topic that certainly catches the attention of experienced software developers looking to extract every bit of performance gain.  It can certainly help optimize programs, especially those that perform specialized work.  It is not a one-size-fits-all solution, though, and as with every complex technology, its use needs to be studied and justified, and its impact on the final application should be monitored and analyzed.
+
+## More information
+
+LINK TO BLOG POST HERE
```
s-makin, line 260: Don't forget to link :D

explanation/performance/perf-pgo.md
sergiodj: Nice.  Done.

explanation/performance/perf-pgo.md
sergiodj: Done.

explanation/performance/perf-pgo.md
sergiodj: I personally don't think so, no.  This will add too much complexity to an already complex document, IMHO :-/.

explanation/performance/perf-pgo.md
sergiodj: ACK, and a fair point.  Done.

sergiodj: Done.

explanation/performance/perf-pgo.md
sergiodj: Good point.  Let me explain this point a bit better.

explanation/performance/perf-pgo.md
sergiodj: Thanks.  I decided to simplify and remove the "silver bullet" part as suggested.

explanation/performance/perf-pgo.md
sergiodj: Fair point.  I've rewritten that part.

explanation/performance/perf-pgo.md
sergiodj: Thanks!  Since the blog post will be available later, I have temporarily removed this subsection and will reintroduce it when I have the final link.

explanation/performance/perf-pgo.md
sergiodj: Fair point.  I will finish the blog post soon and will update the document to include it in the suggested places.

sergiodj:
@cpaelzer @s-makin Thanks for all your feedback.  I have addressed them and force-pushed the results.  I believe this is ready to be published now, but I will let you take another look just in case.

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,226 @@
...
+
+After confirming that everything looks OK, we are ready to start the benchmark.  Let's run the command:
+
+```bash
+$ openssl speed -seconds 60 -evp md5 sha512 rsa2048 aes-256-cbc
+```
+
+This will run benchmark tests for `md5`, `sha512`, `rsa2048` and `aes-256-cbc`.  Each test will last 60 seconds, and they will involve calculating as many cryptographic hashes from `N`-byte chunks of data (with `N` varying from 16 to 16384) with each algorithm (with the exception of `rsa2048`, whose performance is measured in signatures/verifications per second).  By the end, you should see a report like the following:
+
+```bash
```
s-makin, line 51:
```suggestion
```text
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,226 @@
...
+md5             140978.58k   367562.53k   714094.14k   955267.21k  1060969.81k  1068296.87k
+                  sign    verify    sign/s verify/s
+rsa 2048 bits 0.000173s 0.000012s   5776.8  86791.9
+```
+
+### Building OpenSSL for profiling
+
+Before we are able to profile OpenSSL, we need sure that we compile the software in a way that the generated program meets the requirements.  In our case, you can use the `file` command after building the binary to make sure that it has not been stripped:
+
+```bash
```
s-makin, line 65:
```suggestion
```text
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,226 @@
...
+$ file /usr/bin/openssl
+/usr/bin/openssl: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=42a5ac797b6981bd45a9ece3f67646edded56bdb, for GNU/Linux 3.2.0, with debug_info, not stripped
+```
+
+Note how `file` reports the binary as being `with debug_info, not stripped`.  This is the output you need to look for when inspecting your own binaries.
+
+You can use the `eu-readelf` command (from the `elfutils` package) to determine whether any compression was performed with `dwz`:
+
+```bash
```
s-makin, line 74:
```suggestion
```text
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,226 @@
...
+[...]
+```
+
+If the `Flags` field has `C` (as in `compressed`) in it, this means that `dwz` was used.  You will then need to figure out how to disable its execution during the build; for this particular build, this was done by overriding the `dh_dwz` rule on `debian/rules`, setting it to an empty value, and executing the build again.  The particularities of how to skip build steps depend on the build system being used, and as such this guide will not go into further details.
+
+### Using `perf` to obtain profile data
+
+With OpenSSL prepared and ready to be profiled, it's now time to use `perf` to obtain profiling data from our workload.  First, let's make sure we can actually access profiling data in the system.  As `root`, run:
+
+```bash
```
s-makin, line 90:
```suggestion
```text
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,226 @@
...
+
+Note that the only thing that differs between the Intel and AMD variants is the PMU event to be monitored.  Also, note how we are using `sudo` to invoke `perf record`.  This is necessary in order to obtain full access to Linux kernel symbols and relocation information.  It also means that the `openssl-nonpgo.perfdata` file ownership will need to be adjusted:
+
+```bash
+$ sudo chown ubuntu:ubuntu openssl-nonpgo.perfdata
+```
+
+Now, we need to convert the file to `gcov`.
+
+#### Alternative scenario: profiling the program using other approaches
```
s-makin, line 128: Looking at the rendered version, I'm a bit confused why this is so indented - since it's not really part of the first subsection, should it be promoted to a level 2 header? (##)

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,226 @@
...
+	--binary /usr/bin/openssl \
+	--gcov openssl-nonpgo.gcov \
+	--profile openssl-nonpgo.perfdata \
+	--gcov_version 2 \
+	-use_lbr false
+```
+
+`create_gcov` is verbose and will display several messages that may look like something is wrong.  For example, the following output is actually from a successful run of the command:
+
+```
```
s-makin, line 164:
```suggestion
```text
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,226 @@
...
+[WARNING:[...]/perf_reader.cc:1069] Skipping unsupported event PERF_RECORD_EVENT_UPDATE
+[WARNING:[...]/perf_reader.cc:1069] Skipping unsupported event PERF_RECORD_CPU_MAP
+[WARNING:[...]/perf_reader.cc:1069] Skipping unsupported event UNKNOWN_EVENT_17
+[WARNING:[...]/perf_reader.cc:1069] Skipping unsupported event UNKNOWN_EVENT_18
+[...] many more lines [...]
+```
+
+Nevertheless, you should inspect its output and also make sure to check the `$?` shell variable to make sure that it has finished successfully.  However, even if `create_gcov` finishes with exit `0` the generated file `openssl-nonpgo.gcov` might not be valid.  It is also a good idea to use the `dump_gcov` tool on it and make sure that it actually contains valid information.  For example:
+
+```bash
```
s-makin, line 176:
```suggestion
```text
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,226 @@
...
+-fauto-profile=/path/to/openssl-nonpgo.gcov
+```
+
+Make sure to adjust the path to the `openssl-nonpgo.gcov` file.
+
+Also note that enabling PGO will make GCC automatically enable several optimization passes that are usually disabled, even when using `-O2`.  This can lead to new warnings/errors in the code, which might require either the code to be adjusted or the use of extra compilation flags to suppress these warnings.
+
+After the software is built, you can install this new version and run the benchmark tests again to compare results.
+
+### Our results
```
s-makin, line 210:
```suggestion
## Our results
```

explanation/performance/perf-pgo.md
```diff
@@ -0,0 +1,226 @@
...
+
+* Try to profile your software in a regular, production-like setup.  For example, don't try to artificially minimize your system's background load because that will produce results that will likely not reflect workloads found in the real world.  On the other hand, be careful about non-representative background load that might interfere with the measurements.
+
+* Consider whether you are dealing with code that has already been highly optimized.  In fact, this is exactly the case with our example: OpenSSL is an ubiquitous cryptographic library that has received *a lot* of attention from several developers, and it is expected that cryptographic libraries are extremely optimized.  This is the major factor that explains why we have seen minimal performance gains in our experiment: the library is already pretty performant as it is.
+
+### Conclusion
```
s-makin, line 224:
```suggestion
## Conclusion
```

s-makin (commented):
LGTM :) thanks for all your work on this!

I have taken the liberty of adding a few suggestions on adjusted header levels, and also I noticed in the rendered RTD page that the bash code blocks (when applied to the output) lead to some weird and distracting colours being applied. You can either split up the command + output, or change the code block language to "text" to avoid this. I'll leave it up to you to decide which you prefer :)

I'll let @cpaelzer take the approving review, I'm happy for it to be merged once he's happy.

explanation/performance/perf-pgo.md
sergiodj: Hm, the problem with promoting this to a level 2 header is that it breaks the flow of instructions.  Let me see if I can move this subsection down.

s-makin (approved):
LGTM :)

explanation/performance/perf-pgo.md
sergiodj: Done.

explanation/performance/perf-pgo.md
sergiodj: Done.

PR #57, opened 2024-09-26
Add new tutorial on managing software

s-makin:
This tutorial, aimed at newer Ubuntu users, should provide a comprehensive walkthrough of
managing software, which explains many of the terms and concepts they need to know
to use the rest of the content of the server docs and help them be successful.

Note: I'm having issues with Multipass currently, so the later part of the tutorial I
haven't been able to properly test to ensure that it works on a fresh VM


tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+
+```bash
+sudo apt upgrade
+```
+
+The `upgrade` command looks at the metadata in the package index we just
+updated, finds the packages with available upgrades, and lists them for us. It
+also asks us to confirm if we want to do the upgrade. Let's type <kbd>Y</kbd>
+to confirm that we want to, and the upgrade will proceed. This may take a few
+minutes.
```
YanisaHS, line 154: You could also say they can press Enter for the default (Yes in this case because Y is capitalized)

I also think you should have this section start with "This command will ask you to confirm..." upfront. Not having that information be the first thing that came after the `upgrade` command is confusing. Or, you could inform the user before they run it that they'll be asked a prompt

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+In many places, you will see reference to `apt-get` and `apt-cache` instead of
+`apt`. Historically, the *database* part of APT was accessed using `apt-cache`
+(e.g. `apt-cache show ipcalc`), and the *packages* part of APT used `apt-get`
+(e.g. `apt-get install ipcalc`).
+
+APT has recently been streamlined, so although it uses `apt-get` and
+`apt-cache` "behind the scenes", you don't need to worry about remembering
+which command to use -- you can use `apt` directly. To find out more,
+refer to the manual pages for these commands. Type `man apt`, `man apt-get` or
+`man apt-cache` in your terminal to access them.
```
YanisaHS, line 254: Should you give a quick mention what man pages are?

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+
+## Installing deb packages
+
+For the following examples, we're going to use two popular webserver packages:
+Apache2 and nginx. However, we won't install anything just yet -- let's just
+take a look at the commands involved to understand what's happening.
+
+Installing deb packages using APT is done using the `apt install` command. We
+can install a single package:
```
YanisaHS, line 263: This part (implicitly) guides the user to actually run this command though. But if they install Apache here, then they won't be able to follow the next section, which assumes they don't have Apache installed yet.

Can you pick a different package for these two examples with apache?

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+
+Or multiple packages at once, by including them in a space-separated list like
+this:
+
+```bash
+sudo apt install apache2 nginx
+```
+
+### Dependencies
+
```
YanisaHS, line 277: This section made me think - maybe you should show how to install and uninstall packages right at the beginning of the tutorial? So the user is equipped with that knowledge in case they get off track in your tutorial. It could be done as a quick overview of some commands they might need along the way (or something like that)

There's a lot of picky steps here, and if you accidentally install something you shouldn't and get off, it's difficult to get back (well if your user isn't that familiar with Linux at least, which might be the case for this tutorial)

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+Depends: apache2-bin (= 2.4.58-1ubuntu8.4), apache2-data (= 2.4.58-1ubuntu8.4), apache2-utils (= 2.4.58-1ubuntu8.4), media-types, perl:any, procps
+Recommends: ssl-cert
+Suggests: apache2-doc, apache2-suexec-pristine | apache2-suexec-custom, www-browser, ufw
+[...]
+```
+
+In Ubuntu, the default configuration of `apt install` is set to install
+recommended packages alongside `depends`, so when we ran the
+`apt install apache2` command, `ssl-cert` was included in the packages
+to be installed (even though it's only recommended, and not strictly needed).
```
YanisaHS, line 362: Maybe include a step to show that ssl-cert is already installed? (`apt list ssl-cert`?)

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+
+We can override this behaviour by passing the `--no-install-recommends` flag to
+our command, like this:
+
+```bash
+sudo apt install apache2 --no-install-recommends
+```
+
+Then the output becomes the following (let's type <kbd>n</kbd> at the prompt
+again to avoid installing for now):
```
YanisaHS, line 372: Can you note how to uninstall apache2 if they accidentally already clicked Y? Otherwise, they won’t be able to follow any of this part of the tutorial

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+
+Most of a package's configuration is handled through
+[configuration files](https://www.debian.org/doc/debian-policy/ap-pkg-conffiles.html#automatic-handling-of-configuration-files-by-dpkg)
+(often known as **conffiles**).
+
+Package conffiles are different from all other files delivered in a package.
+A package may have any number of conffiles (including none!). Conffiles are
+explicitly marked by the package maintainer during development to protect them
+from being automatically upgraded, whilst all other types of files are not --
+any changes you make to regular files *will be overwritten* during an upgrade.
```
YanisaHS, line 626: Doesn’t it prompt them if they want to override their existing conf files though? Do you want to note that? (I’m not actually sure if they’re always prompted–Landscape Server prompts if they want to overwrite configuration files)

YanisaHS, line 626: I think it'd be good to provide an example of information that could be in conf files. The user is really thrust into conffile world without much context here (aside from the official deb doc)

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+although they are often found in the `/etc/` directory, they don't *have* to be
+there. As we saw before, the only thing conffiles have in common is that the
+package maintainer decided to mark them as such.
+
+But that's our clue! So once more, `dpkg` can come to our rescue. The
+following command will show us the subset of files in a package that have been
+marked as "conffiles" by the maintainer:
+
+```bash
+dpkg-query --show -f='${Conffiles}\n' apache2
```
YanisaHS, line 660: Can you explain what the parts of this command are doing? So it’s not like “hello beginner just run this magic”

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+
+```bash
+sudo sed -e 's/LogLevel warn/LogLevel debug/' -i /etc/apache2/apache2.conf
+```
+
+Then, we'll restart our Apache2 server so that we can activate our configuration
+changes:
+
+```bash
+sudo systemctl restart apache2
```
YanisaHS, line 711: After running this, I got `Failed to restart apache2.service: Unit apache2.service not found.`

I don't think the tutorial actually has us reinstall apache2, so that should be added (unless I just missed it)

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+By using the `--reinstall` flag, we can force `apt` to re-unpack all of the
+default content. If we then verify once more...
+
+```bash
+dpkg --verify apache2
+```
+
+...we can see that our change to the conffile has been preserved because
+`dpkg` the checksums are different, but the `a2enmod` file isn't
+listed anymore because it has been restored to the default. Phew!
```
YanisaHS, line 787: I think you should put the expected output before this.

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+
+```{note}
+We can use `sudo apt install <package>` to upgrade an installed package, but
+this will only upgrade to the latest version. In our case, we were already on
+the latest version of Apache2, so we needed to force APT to re-unpack the
+content to overwrite our "silly" changes.
+```
+
+## Removing packages
```
YanisaHS, line 796: I like this, but I think you should have a little bit of this at the top in case the user accidentally installs something and gets off track from your tutorial. You could still have this section here but reword it slightly

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+  include additional security coverage from Canonical under
+  [Ubuntu Pro](https://ubuntu.com/pro), which is available free for personal
+  use on up to five machines.
+- **restricted** contains the packages that are officially supported by
+  Canonical but are not available under a completely free license.
+- **multiverse** contains community-maintained proprietary software -- these
+  packages are completely unsupported by Canonical.
+
+Every Ubuntu release (`noble`, `jammy`, etc) is also split into **pockets**
+which are related to the development/release lifecycle:
```
YanisaHS, line 1110:
```suggestion
Every Ubuntu release (`noble`, `jammy`, etc) is also split into **pockets** (sections of packages)
which are related to the development/release lifecycle:
```

Or similar? Your doc doesn't really note what pockets are. It defines them officially, but it doesn't clear up what they actually are to a beginner

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+There are a couple of dependencies that we'll want to install that don't come
+packaged with the development version, so let's install those now:
+
+```bash
+sudo apt install autoconf libtool-bin
+```
+
+Next we'll want to configure the source tree -- the prefix is the most
+important option, so let's set that (for the sake of this example, we'll just
+set it to the default location):
```
YanisaHS, line 1232: why do we want to configure the source tree? What is the source tree and why does it need to be configured? (You lightly explain below, but I think context should be provided before they run the command)

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+
+```bash
+./configure --prefix=/usr/local/apache2
+```
+
+Configuration is an important step, because this is where the Makefiles are
+created that will allow us to install the package. It will take a few moments
+as it checks all the files.
```
YanisaHS, line 1240: The user is thrown into make/Makefiles here too. Can you provide context?

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+means that there's no easy way to uninstall or update the package in the
+future since the package managers (`apt` and `dpkg`) can't track them. So one
+thing we can do is to create an installable deb package (which will help to
+make removal easier!) -- to do that, we can use the `fpm` utility.
+This is a Ruby package that needs to be installed via `gem`:
+
+```bash
+sudo apt install ruby ruby-dev gcc
+gem install fpm
+```
```
YanisaHS, line 1260: I’m not a fan of using Ruby as the example for a couple reasons:
1. Ruby isn’t very popular anymore, so some users (particularly younger ones who weren’t connected to tech during the Ruby on Rails time) might not be familiar with it, and a more general example would be less distracting
2. Ruby’s package manager is called with `gem`, but Ruby also calls its packages “gems”, so it’s difficult for people brand new to this stuff to even Google what “ruby” or “gem” is

You could use an example with `pip` instead?

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+
+Now at this point, we could consider ourselves done, but installing this way
+means that there's no easy way to uninstall or update the package in the
+future since the package managers (`apt` and `dpkg`) can't track them. So one
+thing we can do is to create an installable deb package (which will help to
+make removal easier!) -- to do that, we can use the `fpm` utility.
+This is a Ruby package that needs to be installed via `gem`:
+
+```bash
+sudo apt install ruby ruby-dev gcc
```
YanisaHS, line 1258: Prompts you to Y/N to continue

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+gem install fpm
```
YanisaHS, line 1259:
```suggestion
sudo gem install fpm
```

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+```
+
+Then we can create the package:
+
+```bash
+fpm -s dir -t deb -n apache2-custom -v 2.4.62 /usr/local/apache2
```
YanisaHS, line 1265: can you give context to what’s in this command? (say what all the flags are/etc).

I also got this error:
```bash
Invalid package configuration: Cannot package the path './usr/local/apache2', does it exist? {:level=>:error}
```

YanisaHS (approved):
This tutorial is insane but also awesome!!! My review is done :hot_face: 

("Approving" because I accidentally made this a formal review instead of just single comments ...)

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+After all, Ubuntu is endlessly customisable according to your needs.
+
+To help you get the most from your Ubuntu experience, this tutorial will walk
+you through managing the software on your Ubuntu machine. Although we'll be
+using Ubuntu Server as our example, this tutorial is equally applicable to the
+Ubuntu Desktop.
+
+There are two types of software found in Ubuntu: **Debian packages** and
+**snaps** -- we will learn about both!
+
```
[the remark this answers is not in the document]
s-makin: Added a prerequisites section :)

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+to launch our server and run the commands.
+
+Multipass is great for quickly creating Ubuntu virtual machines, so we will use
+that. On Ubuntu, you can install Multipass by running the following command in
+your terminal:
+
+```bash
+sudo snap install multipass
+```
+
```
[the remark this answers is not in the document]
s-makin: I'm wrestling with this a bit. It doesn't feel like the right place to talk about it, at this point the sudo is just part of the vehicle to get to the tutorial, and there are no good up-to-date or reliable resources to point to. I might find a place to drop in a note later on -- someone who's not a complete beginner will know about sudo already, and a complete beginner might not even realise the significance. Or I could sidestep it by giving a link to the snapstore page in the instructions.

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+Multipass will download the most recent daily **image** and create the VM for
+us. It may take a little time, depending on the speed of your internet
+connection.
+
+The **image** is the collection of files we need to install and run Ubuntu. We
+don't need to specify "server" or "desktop" anywhere in our command, because
+the image is the same for both. The only difference between Ubuntu Server and
+Ubuntu Desktop is the subset of software packages we use from the image - we
+will see this later!
```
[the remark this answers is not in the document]
s-makin: Great point! Added

tutorial/managing-software.md
```diff
@@ -0,0 +1,1394 @@
...
+of all the default software. 
+
+Debian packages, commonly referred to as **"debs"** are the standard software
+in Ubuntu. They can be identified by the `.deb` file extension and usually come
+in a pre-compiled binary format (so you don't need to compile it yourself).
+
+Your system software (and other deb packages) are handled through the
+[Advanced Packaging Tool](https://ubuntu.com/server/docs/package-management#advanced-packaging-tool)
+(APT). APT is a package *repository*; it contains both the **database** of the
+packages available in Ubuntu, and it contains the **packages** themselves.
```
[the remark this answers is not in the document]
s-makin: I'm not sure that `pip` use is extensive/widespread enough to make the comparison meaningful (especially because APT is the default in Ubuntu), but I can/should probably make the package management aspect more explicit. I'll have a fiddle and see what improvements I can make.

tutorial/managing-software.md
s-makin: Reorganised the contents a bit to hopefully make the "you need to confirm" part a bit more visible, and added the Enter tip (I didn't know about that one!)

tutorial/managing-software.md
s-makin: Hmm...yes, probably! And also a link to the online ones. Thanks for the reminder :)

tutorial/managing-software.md
s-makin: I think that's a great point. I don't think my "we won't install anything yet" is strong enough given that a user may just start running commands on autopilot. I think I'll generalise these commands so that they just won't install anything.

s-makin: On reflection, this also seemed like the ideal time to talk about sudo, so added that content here

tutorial/managing-software.md
s-makin: If they're following the commands/instructions exactly, a user shouldn't have ssl-cert installed yet. This whole section is just about the different dependency types and how to find out what's recommended/required/suggested etc (and a bit of an introduction in how to use flags).

s-makin: I need to make it a bit stronger in the intro to the section I think.

tutorial/managing-software.md
s-makin: Good point, I missed that. I've added a bit on what sort of info *might* be in a package (since it depends on the package).

tutorial/managing-software.md
s-makin: Not easily/in a way that doesn't invite more questions, but I've tried :upside_down_face:. I've added the bits of the command to the relevant keywords of the paragraph above to highlight them, but it's still not ideal.

tutorial/managing-software.md
s-makin: Do you know, you're completely right. In all the previous steps I forgot to add that in! Great catch

tutorial/managing-software.md
s-makin: I've also added the manpage command, partly so users can investigate the options if they want, but also to get them used to using the manpages. I'm still not 100% happy, but I think it's better than what I originally had.

tutorial/managing-software.md
s-makin: Fixed

tutorial/managing-software.md
s-makin: That's fair! :) added

tutorial/managing-software.md
s-makin: This is where everything has the potential to become very confusing lol. The thing that really is "sections" of packages is the components, where they're clearly divided into one or the other. You can't have a package in both main and universe afaik. Pockets aren't sections of packages, so much as sections of components. I've never been able to find a good, non-confusing explanation, so I'll try and write one.

s-makin: I've added a mermaid diagram as well, this may help a bit

tutorial/managing-software.md
s-makin: In general, I'm ok with providing exposure to older tech (we even have content about Ruby in the server docs). In this case, we can't get `fpm` through `pip` so I'm a bit tied to Ruby here :/ (the alternatives you can get through pip are even more fiddly and annoying to use, so this is the least-worst option imo)

tutorial/managing-software.md
YanisaHS: FWIW I described pockets as "sections" and components as "categories" and made a diagram here: https://ubuntu.com/landscape/docs/explanation-about-repository-mirroring. It's def still confusing, but just passing the link along if it helps to see a second opinion on how that stuff could be explained :blush:

tutorial/managing-software.md
s-makin: Context added!

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+they are no longer maintained, or conversely, are the newest version still in
+development!
+
+We can handle deb packages that aren't in the APT repository using ``dpkg`` --
+all we need is to download the deb, and we can run a command like this to
+install it:
+
+.. code-block:: bash
+
+    sudo dpkg -i <package name>
```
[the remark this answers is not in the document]
cpaelzer: If someone wants to do the very same BUT with also automatically handling dependencies to be resolved from the archive, they can use `apt install ./filename`. You need leading / or ./ to make sure apt knows it is not a package name from the repo. But that allows to install a .deb with the simplicity of if it has dependencies that can be resolved - they would.

Therefore I'd prefer to suggest that. WDYT?

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+
+However, it is also possible to obtain software by building it from the source.
```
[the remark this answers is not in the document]
cpaelzer: Furthermore this whole topic is so very different based on which software you talk about - that none of the instructions might match when someone does something else. So even when moving elsewhere, you should add even more intense "this will be very different depending on what you do, this is just one example"

And again, not good to be here in the tutorial.

Makes the otherwise great "how you should use it" complicated end ends with showing something people should better not use - at least not until they really++ know what they do.

Moving installing from manually downloaded .deb
And building source to separate pages each and linking from the tutorial as "there is more, but you should almost never need that and it comes with drawbacks to maintainability of the systems (no updates, ...)" is what I'd expect.

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+* How to see if package files have been changed:
+  * ``dpkg --verify <package>``
+
+* ...And if a non-conffile has been changed by accident, we can fix it with:
+  * ``sudo apt install --reinstall <package>``
+
+* We know that our changes to conffiles are always safely preserved, while
+  changes to non-conffiles are reverted at the next upgrade or security fix.
+
+* Importantly, we know how to verify checksums with ``md5sum`` or similar
```
cpaelzer, line 1607: You did not mention the build from source here, which is good - I do not want it here :-)

But this list also forgets about the install from a .deb file - which ok'ish.
I'm unsure if that should be here or also in a page of its own with big disclaimers (essentially the third party explanation we have).

cpaelzer (changes requested):
Already quite good, I've left hopefully helpful suggestions.
But even more so at the end there was some content that I disagree (it should not be part of the tutorial).

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
+.. _managing-software:
+
+Managing your software
+**********************
+
+If you are new to Ubuntu, you may be wondering what to do after installation.
+After all, Ubuntu is endlessly customisable according to your needs.
+There are two types of software found in Ubuntu: **Debian packages** and
```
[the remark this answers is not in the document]
s-makin: In a how-to I would, but here I don't want to encourage anyone to skip ahead over the setting up a VM part.

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+
+    multipass launch noble --name tutorial
+
+Multipass will download the most recent daily **image** and create the VM for
+us. It may take a little time, depending on the speed of your internet
+connection.
+
+An Ubuntu **image** is a collection of files we need to install and run Ubuntu.
+We don't need to specify "server" or "desktop" anywhere in our command, because
+the image is the same for both. The only difference between Ubuntu Server and
```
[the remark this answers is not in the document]
s-makin: Definitely worth thinking about. I think if we wanted to expand the tutorial with a more cloud-based topic that might make a really good extension to it. For now I wanted to try and give a soft landing just to introduce the terminology and (hopefully) make clear that there's not a massive difference between Server/Desktop. That was the thinking behind the repetition anyway. I know a lot of Desktop users are likely to end up using this tutorial, so I want them to feel "safe" to do so.

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+Ubuntu Desktop is the subset of software packages we use from the image - we
+will see this later!
+
+Now we can access the VM by running:
+
+.. code-block:: bash
+
+    multipass shell tutorial
+
+We will get a "Welcome to Ubuntu 24.04 LTS" message. Notice that when we run
```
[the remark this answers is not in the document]
s-makin: Good idea, I've removed it and just left "Welcome to Ubuntu"

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+    ubuntu@tutorial
+
+This shows that we are inside the VM, and this is where we will run all our
+commands.
+
+Updating the system with APT
+============================
+
+The first thing we always want to do with a new system (whether a VM or a
+"live" machine) is to run an update to make sure we have the latest versions
```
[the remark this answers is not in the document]
s-makin: That's much better, thanks!

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+of all the default software.
```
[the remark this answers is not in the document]
s-makin: went with "pre-installed" - thanks

tutorial/managing-software.rst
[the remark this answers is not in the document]
s-makin: On reflection I shortened it to "is to make sure we have the latest versions..." and avoid both update and upgrade.

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+commands.
+
+Updating the system with APT
+============================
+
+The first thing we always want to do with a new system (whether a VM or a
+"live" machine) is to run an update to make sure we have the latest versions
+of all the default software. 
+
+Debian packages, commonly referred to as **debs**, are the standard software
```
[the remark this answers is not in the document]
s-makin: I'm happy with that - changed to your suggestion

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+in Ubuntu. They can be identified by the ``.deb`` file extension. 
+
+Every Linux distribution has their own preferred **package manager** for
+installing, updating and removing packages. In Ubuntu, the default package
+manager is `Advanced Packaging Tool <https://wiki.debian.org/AptCLI>`_ (or APT,
+for short).
+
+APT handles all of your system software (and other deb packages). APT is a
```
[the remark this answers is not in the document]
s-makin: I take your point - but that might be a complicated way to describe it. I've tried to come up with an alternative to the original that steers a bit closer to APT just being an interface. Let me know if this is correct or if it still needs a bit more work.

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+    Get:1 http://archive.ubuntu.com/ubuntu noble-updates/main amd64 libopeniscsiusr amd64 2.1.9-3ubuntu5.1 [49.1 kB]
+
+APT combines the various elements; the package name (``libopeniscsiusr``),
+version (``2.1.9-3ubuntu5.1``), source (``noble-updates/main``), etc into a single
+URL that it can use for the download. It then unpacks the package and applies
+the upgrade to the system.
+
+.. note::
+    These commands only upgrade the packages for the release of Ubuntu that we are
+    using (24.04 LTS). If we wanted to upgrade to the next release of Ubuntu, we
```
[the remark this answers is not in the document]
s-makin: Much better. I've also added a parenthetical example to hopefully make it even clearer what we're talking about.

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+The ``dist-upgrade`` command is able to resolve conflicts between package
+versions, but it *could* end up removing some packages -- so although
+``apt upgrade`` is safe to use unattended (in a script, for example), you should
+only use ``dist-upgrade`` when you can pay attention to it.
+
+Searching APT
+-------------
+
+Now we're up-to-date, we can start exploring! As with any other database, we
+can search APT to find software. Let's say that we want to find a webserver,
```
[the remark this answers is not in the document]
s-makin: Since we're using APT to search, I've updated the section heading to "Searching with APT" (hopefully accurate enough to reflect what you're suggesting here) and updated the text with your suggested wording.

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+
+For the examples for this section, we're going to use the popular webserver
+package, `Apache2 <https://httpd.apache.org/>`_.
+
+APT gives us a lot of details about what will be included in the installation,
+and it's always important to understand the implications of a command
+*before* we run it. We'll be taking a close look at the details APT gives us,
+so we need to be careful in this section.
+
+When we run a command that asks us "``Do you want to continue? [Y/n]``", make
```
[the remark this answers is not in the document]
s-makin: I'm not sure. It seems like a helpful thing to mention, but since most guides and instructions a new user runs across will show them sudo commands, my intention was to talk about sudo and encourage users to slow down and read and understand what's presented (and not just spam "enter" through confirmation screens). I don't think I've ever seen the simulate flag before, so it's not commonly used in documentation. I think I'll add it, but I propose to do it at the end of the section. wdyt?

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+About dependencies
+------------------
+
+As we hinted earlier, packages often come with **dependencies** -- other
+packages that *your* package needs so it can function. Sometimes, a package
+might depend on a specific version of another package. If a package has
+dependencies, then installing a package via ``apt`` will also install any
+dependencies, which ensures the software can function properly.
+
+We can see what dependencies a package has using the ``install`` command. Let's
```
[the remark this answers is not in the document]
s-makin: I've tried to reword this a bit without going too far into explaining dependencies. Let me know if it's better or not.

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+Types of dependencies
+^^^^^^^^^^^^^^^^^^^^^
+
+The relationship between a package and any other packages follows the
+`Debian policy on binary dependencies <https://www.debian.org/doc/debian-policy/ch-relationships.html#binary-dependencies-depends-recommends-suggests-enhances-pre-depends>`_,
+which we'll briefly look at here. The most common ones you might come
+across are: ``depends``, ``recommends``, and ``suggests`` (although there are
+others!), so we'll take a look at these three.
+
+- **depends**: Absolutely required, the package won't work without it.
```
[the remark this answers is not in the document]
s-makin: Made the wording a bit stronger in both places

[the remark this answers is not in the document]
s-makin: (in the second place, made into a warning)

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+packages at the time. We might have found another use for ``ssl-cert``, perhaps
+in a script that makes use of SSL certificates. So how can we keep the
+``ssl-cert`` package, even though it's flagged for autoremoval?
+
+We can solve this problem, and un-flag the ``ssl-cert`` package for removal, by
+*manually* installing it:
+
+.. code-block:: bash
+
+    sudo apt install ssl-cert
```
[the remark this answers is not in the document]
s-makin: Agreed and good point - added a little explanation here to that effect. I'm still pondering on where to introduce the --dry-run option. This doesn't feel like the most relevant place for it.

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+`configuration files <https://www.debian.org/doc/debian-policy/ap-pkg-conffiles.html#automatic-handling-of-configuration-files-by-dpkg>`_
+(often known as **conffiles**). Conffiles often contain things like file paths,
+logs and debugging configuration, kernel parameters (which can be changed to
+optimise system performance), access control, and other configuration settings.
+The actual parameters available will vary from one package to another.
+
+Package conffiles are different from all other files delivered in a package.
+A package may have any number of conffiles (including none!). Conffiles are
+explicitly marked by the package maintainer during development to protect them
+from being automatically upgraded, whilst all other types of files are not --
```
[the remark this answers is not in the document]
s-makin: I think that's much better, updated to that.

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+any changes you make to regular files *will be overwritten* during an upgrade.
```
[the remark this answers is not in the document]
s-makin: Good catch - made it more specific

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+understand how such conflicts are handled.
+
+We can show the four possible upgrade scenarios using the following table. What
+happens during an upgrade depends on whether the conffile on our system has
+been changed by us ("changed/not changed by user"), and whether the default
+content has been changed by the package maintainer ("same as/changed from
+default"):
+
+========================== =================== =========================
+The conffile is...         **same as default** **changed from default**
```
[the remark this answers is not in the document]
s-makin: That's a great point. I picked "default" because it made the table more compact but I see the ambiguity I've created by doing that. I've changed it to "changed/not changed by maintainer" -- hopefully this is better!

tutorial/managing-software.rst
```diff
@@ -0,0 +1,1611 @@
...
+  time.
+
+This is why earlier, we saw that some updates came from ``noble-updates`` or
+``noble-security``. These refer to updates and security updates from the noble
+series (respectively). Pockets are usually appended to the end of the series,
+and it's quite common to see the hyphen (``-``) included when referring to
+pockets. 
+
+Remember that the original version of the ``apache2`` package that we saw came
+from ``noble`` -- the ``-release`` pocket only includes the software that was
```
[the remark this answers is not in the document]
s-makin: I didn't realise it would split that, never seen that happen before. I've tried to reword the paragraph a bit to avoid that word appearing at the end of the line, but it will depend on the screen size to some extent. At least on a regular monitor it shouldn't happen anymore.

tutorial/managing-software.rst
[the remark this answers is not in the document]
s-makin: Removed this entire section, as discussed

tutorial/managing-software.rst
s-makin: I don't want to draw too much attention to it, since we prefer people don't do it and are only showing it as an FYI because we're doing it in a sandbox. Actually, now that I reflect more on the reasons to remove all the building from source stuff, I wonder if we should also remove the install from .deb stuff - but it does provide a convenient segue into snaps.

PR #64, opened 2024-10-09
virtiofsd: avoid option not always present

cpaelzer:
The virtiofsd server changed a lot in recent years, the version in Noble is no more having the flock option. I considered doing a huge "if release then" but the benefit is minimal. We already link to upstreams collection of examples which has this and many more which all work depending on your needs and versions.

Let us just take away the flock hint to avoid users of the more modern stack that does it well right away to be blocked by an error about non existing options.


s-makin (approved):
Sounds sensible to me :)

PR #66, opened 2024-10-16
Remove duplicated block of text

Pexers:
Removes duplicated block of text


s-makin (approved):
Great catch! Thank you so much for taking the time to submit a PR, your help in improving the Server docs is much appreciated :)

PR #83, opened 2024-11-19
Fix broken link to Ubuntu bug

daveedoo:
As in the title


s-makin (approved):
Good catch! Thanks for taking the time to fix it :)

PR #90, opened 2024-12-02
Add Rediraffe support

s-makin:
This PR adds support for the [Rediraffe Sphinx extension](https://sphinxext-rediraffe.readthedocs.io/en/latest/).

This means that we don't have to keep track of what moved/deleted/renamed files need redirects (and then have to remember to add those to RTD after the fact). Redirects can be added alongside the changes in the same PR, so they are "live" as soon as the change is made.

### DIFF
--- a/contributing/submit-contribution.rst
+++ b/contributing/submit-contribution.rst
@@ -35,8 +35,8 @@ Check if you need redirects
 ===========================
 
 If you rename, move or delete an existing file, a corresponding redirect must
-be created to ensure users don't run into 404 errors in the published
-documentation.
+be created to ensure users don't run into 404 errors when clicking links in the
+published documentation.
 
 To set up a redirect, add a line to the end of the `redirects.txt` file in the
 root directory, in the following format:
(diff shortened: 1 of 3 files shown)


contributing/submit-contribution.rst
```diff
@@ -29,7 +29,26 @@ You can run:
...
    
 To perform a full spelling and link check. You can also run ``make`` by itself
-to see a list of all the possible ``make`` targets. 
+to see a list of all the possible ``make`` targets.
+
+Check if you need redirects
+===========================
+
+If you rename, move or delete an existing file, a corresponding redirect must
+be created to ensure users don't run into 404 errors in the published
```
bryceharrington, line 38: errors "when clicking links" in the published

redirects.txt
```diff
@@ -0,0 +1,9 @@
+# Comment lines start with a hash (#) and are ignored
+# Each redirect should appear on its own line
+# Do not include a file extension
+# Paths don't need a slash in front of them
+# We are using the dirhtml builder so we need a trailing slash at the end
+# The paths should match what's in the build directory
+# Example:
+# redirect/from redirect/to
```
bryceharrington, line 8: The directions say a trailing slash is necessary at the end, which makes the example here a bit confusing.  I assume what it really meant to say is that a trailing slash is required "for directories"?

The example might be more examplly if it looked more like an actual path in our docs, e.g.:
```
# howto/ how-to/
# how-to/foo.rst how-to/bar.rst
```

bryceharrington (changes requested):
Couple clarifications needed.  Also I think the boilerplate directions in the redirects.txt page could be improved a bit, although I'm sure once there's actual content there it'll be mostly obvious how to add more.

redirects.txt
s-makin: Thanks for picking this up - I very stupidly didn't include the trailing slashes in the example. If you put file extensions it will throw errors. Since we're using dirhtml to build, the built docs look like `/how-to/foo/index.html`, and since every "file" is index.html, you don't need to specify it. The file name is essentially nothing more than another directory. I'll tidy up the example and see if I can make this a bit more explicit to reduce potential for confusion.

bryceharrington: Looks great!

PR #91, opened 2024-12-10
Change install-nvidia-drivers.md to have correct --gpgpu position order

MitchellAugustin:
--gpgpu is ignored if it comes after the verb, since it is erroneously interpreted as a driver version in that case. Instead, it should go before the verb.


cpaelzer:
Thank you @MitchellAugustin , makes sense, is small and good without e.g. massive wording changes that needs further evaluation -> merging

Issue #93, opened 2024-12-11
docs: SSH not restarting with "sudo systemctl restart sshd.service"


WJ4IoT:
I notice the command "sudo systemctl restart sshd.service"" does not work on my system:
and "Failed to restart sshd.service: Unit sshd.service not found."

where the command "sudo systemctl restart ssh.service" just works.

I do not have an original "/etc/ssh/sshd_config" anymore because i am working on an older system but this system is now updated to the latest Ubuntu LTS server version (Ubuntu 24.04.1 LTS (GNU/Linux 6.8.0-50-generic x86_64), I do have "/etc/ssh/sshd_config.ucf-dist"),

How do I become in this case a fresh copy of "/etc/ssh/sshd_config"?
---
*Reported from: https://documentation.ubuntu.com/server/how-to/security/openssh-server/*

cpaelzer:
Hi @WJ4IoT ,
I first wanted to say that seems like a config issue, but it might be legit the more I look into it.

I noticed it was good earlier (e.g. jammy) and later (e.g. plucky) in a quick re-check.
While looking if I should file that bug I spotted that I'm about a month too late => https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/2087949

It is already being fixed (that is why it works in plucky) and on the way for the fix to be backported.

For the docs, we could just make it `sshd`->`ssh` to avoid that issue immediately. And while `ssh.service` isn't as clear to be "the daemon" it is the primary name of the service file - so that should be fine.

WJ4IoT:
Thanks for your swiftly answer, I missed the "fantastic Ubuntu Summit" despite it was on walking distance from of my home.

PR #94, opened 2024-12-12
ssh: fix service reference

cpaelzer:
The old used to work and will work again after [1] is fixed, but the ssh service on disk is /usr/lib/systemd/system/ssh.service and sshd only an alias (which currently happens to be broken on noble). Using ssh.service in our examples avoids readers to stumble until fixed in an SRU and can stay as-is afterwards.

[1]: https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/2087949

Fixes: #93

### DIFF
--- a/how-to/security/openssh-server.md
+++ b/how-to/security/openssh-server.md
@@ -53,7 +53,7 @@ Banner /etc/issue.net
 After making changes to the `/etc/ssh/sshd_config` file, save the file. Then, restart the `sshd` server application to effect the changes using the following command:
 
 ```bash
-sudo systemctl restart sshd.service
+sudo systemctl restart ssh.service
 ```
 
 > **Warning**:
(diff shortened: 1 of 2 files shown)


slyon (approved):
Yes, I can confirm this.

On Plucky:
```
root@pp:~# ll /usr/lib/systemd/system/ssh*.service
-rw-r--r-- 1 root root 561 Oct 30 14:33  /usr/lib/systemd/system/ssh.service
lrwxrwxrwx 1 root root  11 Nov 19 17:06  /usr/lib/systemd/system/sshd.service -> ssh.service
-rw-r--r-- 1 root root 341 Oct 30 14:33 '/usr/lib/systemd/system/sshd@.service'
```

On Noble:
```
$ ll /usr/lib/systemd/system/ssh*.service 
-rw-r--r-- 1 root root 538 Apr  4  2024 /usr/lib/systemd/system/ssh.service
```

lgtm.

PR #95, opened 2024-12-13
Add security intro

s-makin:
Adds an expanded introduction to security and a "suggestions" page that signposts all our other security-related content

(please don't merge yet, still not quite finished adding context to the suggestions page)

### DIFF
--- a/explanation/intro-to/security.md
+++ b/explanation/intro-to/security.md
@@ -2,168 +2,78 @@
 # Introduction to security
 
 Security should always be considered when installing, deploying, and using any
-type of computer system. Although a fresh installation of Ubuntu is relatively
-safe for immediate use on the Internet, it is important to have a balanced
-understanding of your system's security posture based on how it will be used
-after deployment.
+Ubuntu system. Although a fresh installation of Ubuntu is relatively safe for
+immediate use, it is important to have a balanced understanding of your
+system's security posture based on how it will be used after deployment. It's
+important to take a layered approach so that your system's security is not
+dependent on a single 
 
-## General guidance for all systems
+## Server security guidance
 
-It is generally good practice to take a proactive and layered approach to
-security. These suggestions are good practices for any Ubuntu system.
+Since Ubuntu is so endlessly customizable, a full guide to security hardening
+is beyond the scope of this documentation.
+However, there are good practices and security-related packages that could be
+applied to almost any Ubuntu system. See our
+{ref}`security suggestions <security-suggestions>` page for an overview of the
+good habits and practices that can be adopted by anyone running an Ubuntu
+system to make it more secure. It's not necessary to apply every suggestion --
+and the list is not exhaustive by any means -- but each one used creates an
+extra layer of security.
 
-## Keep your system up-to-date
+In a more advanced or complex setup, you may need to go further in your
+security outlook. There are specific packages available for your Server
+that will help with this, and we suggest some in the
+{ref}`advanced security <advanced-security>` section that you might want to
+consider for your use-case. Again, the list is not intended to be exhaustive,
+but rather a starting point.
 
-1. **Regularly update** your Ubuntu Server to keep it protected from known
-   vulnerabilities:
-
-   ```bash
-   sudo apt update && sudo apt upgrade
-   ```
-
-1. **Use the `unattended-upgrade` package** to automatically fetch and install
-   security updates and bug fixes:
-
-   ```
-   sudo apt install unattended-upgrades
-   ```
-
-   By default, `unattended-upgrade` runs daily, but this can be configured. See
-   the `unattended-upgrade`
-   [manual page](https://manpages.ubuntu.com/manpages/noble/en/man8/unattended-upgrades.8.html)
-   for details.
-
-1. **Use Ubuntu Pro**. If you are on an older Ubuntu release and don't want to
-   upgrade just yet, Ubuntu Pro provides
-   [10 years of security coverage](https://ubuntu.com/security/esm) for all
-   packages in the Main and Universe repositories. It is free for personal or
-   business use on up to 5 machines.
-
-1. **Remove packages you don't need**, to minimise the
-   potential attack surface. See our article on
-   {ref}`Package management <package-management>` for more details.
-
-1. It's also a good idea to **use the most up-to-date release** of Ubuntu
-   Server. We have instructions on
-   {ref}`how to upgrade your release <upgrade-your-release>`.
-
-## Access Control
-
-1. **Use and enforce** the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege):
-
-   * This means creating non-root user accounts with as few privileges as possible
-   * Not using `sudo` (root access) except for administration tasks
-   * For more details on basic access control, see our {ref}`guide on user management <user-management>` 
-
-1. Use the **Secure Shell (SSH)** protocol to secure remote access. In Ubuntu,
-   this is managed through the OpenSSH software. For details on setting up
-   OpenSSH, refer to our {ref}`guide to OpenSSH <openssh-server>`.
-
-1. For larger or more complex setups, access control is usually gated through
-   tools such as OpenLDAP. Refer to our
-   {ref}`introduction to OpenLDAP <introduction-to-openldap>`
-   for more details, or see our section
-   {ref}`on how to set up OpenLDAP <how-to-openldap>`.
-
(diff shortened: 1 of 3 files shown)


explanation/intro-to/security.md
```diff
@@ -1,14 +1,79 @@
 (introduction-to-security)=
 # Introduction to security
 
-Security should always be considered when installing, deploying, and using any type of computer system. Although a fresh installation of Ubuntu is relatively safe for immediate use on the Internet, it is important to have a balanced understanding of your system's security posture based on how it will be used after deployment.
+Security should always be considered when installing, deploying, and using any
+Ubuntu system. Although a fresh installation of Ubuntu is relatively safe for
+immediate use, it is important to have a balanced understanding of your
+system's security posture based on how it will be used after deployment. It's
+important to take a layered approach so that your system's security is not
+dependent on a single
```
cpaelzer, line 9: single what?

explanation/intro-to/security.md
```diff
@@ -1,14 +1,79 @@
...
+extra layer of security.
+
+In a more advanced or complex setup, you may need to go further in your
+security outlook. There are specific packages available for your Server
+that will help with this, and we suggest some in the
+{ref}`advanced security <advanced-security>` section that you might want to
+consider for your use-case. Again, the list is not intended to be exhaustive,
+but rather a starting point.
+
+For a more thorough treatment of security in Ubuntu, we recommend checking
```
cpaelzer, line 30: "we also recommend" otherwise it sounds like read here is you ar not serious but there if you are.

cpaelzer: And sadly - as we know and as of today - there people find mostly marketing and not so much actionable

explanation/intro-to/security.md
```diff
@@ -1,14 +1,79 @@
...
+out the [Ubuntu Security documentation](https://ubuntu.com/security).
+
+## Ubuntu Pro
+
+Canonical offers security, compliance and support services through the
+[Ubuntu Pro](https://ubuntu.com/pro) subscription. Ubuntu Pro is available
+for free on up to 5 machines (for business or personal use). Although the
+compliance and certification features of Ubuntu Pro are likely to be of more
+interest to enterprise users, the security patching features are great for
```
cpaelzer, line 39: That sounds odd, how about "security patching features" -> "enhanced security coverage"

explanation/intro-to/security.md
```diff
@@ -1,14 +1,79 @@
...
+anyone using Ubuntu.
+
+All of the Ubuntu Pro features can be managed on the command line via the
+[Ubuntu Pro Client](https://canonical-ubuntu-pro-client.readthedocs-hosted.com/en/latest/)
+utility, which also has an API for easier automation.
+
+### Vulnerability management
+
+In a standard Ubuntu LTS release, security support is provided for packages in
+the Main repository for 5 years. With Ubuntu Pro, this is expanded to 10 years,
```
cpaelzer, line 49: is this mentioning of main worth a link to the page explaining the components?

explanation/security/security_suggestions.md
```diff
@@ -0,0 +1,177 @@
...
+   * For more details on basic access control, see our {ref}`guide on user management <user-management>`.
+
+### Network security
+
+1. **Use a firewall**. In Ubuntu, the uncomplicated firewall (`ufw`) tool is
+   used to configure firewalls. `ufw` is a wrapper around the `iptables` utility
+   (which experienced system admins may prefer to use directly). To get started
+   with `ufw`, check out our {ref}`firewalls` guide.
+
+1. **Use the Secure Shell (SSH)** protocol to secure remote access. In Ubuntu,
```
cpaelzer, line 83: Using ssh is fine, the art is doing it well.

We sadly lack content here on the pages to suggest proper key types and sizes, use passphrases, potentially suggest storing the private key securely as well as something like two factor like http://www.justgohome.co.uk/blog/2013/07/better-two-factor-ssh-authentication-on-ubuntu.html
If not there yet we at least should track that as an extension of the content.

explanation/security/security_suggestions.md
```diff
@@ -0,0 +1,177 @@
...
+   implemented through OpenLDAP. Refer to our
+   {ref}`introduction to OpenLDAP <introduction-to-openldap>`
+   for more details, or see our section
+   {ref}`on how to set up OpenLDAP <how-to-openldap>`.
+1. **Kerberos** is a network authentication protocol that provides identity
+   verification for distributed environments, commonly used in enterprise
+   systems. Learn more in our
+   {ref}`introduction to Kerberos <introduction-to-kerberos>`, or see our
+   section on how to {ref}`set up and use Kerberos <how-to-kerberos>`.
+1. **System Security Services Daemon (SSSD)** is a collection of daemons that
```
cpaelzer, line 118: To be fair to all options, should we include integration into an active directory in this list?

explanation/security/security_suggestions.md
```diff
@@ -0,0 +1,177 @@
...
+   * {ref}`How to set up WireGuard VPN <how-to-wireguard-vpn>`
+
+1. **OpenVPN**
+   
+   * {ref}`About OpenVPN clients <openvpn-client-implementations>`
+   * {ref}`How to install OpenVPN <install-openvpn>`
+
+### Security of communications
+
+1. **TLS/SSL** for secure communication
```
cpaelzer, line 140: I think we need here a short "why you'd want transport security" in general, and can then refer to those examples - because in reality it is always different and depends on the solution they are setting up.

Remiding users to use some transport level secuirty is fair without explaining all potential options.
Just be open that many stacks offer it, some even try to use it by default, but even then it requires certificate/secrets management which is why it almost always needs the system administrator to get it to be really good.

cpaelzer (commented):
Provided some in-line comments to further guide you.
You said it isn't yet fully ready, so I'm neither approving nor rejecting at this stage.

explanation/intro-to/security.md
s-makin: oops, good catch!

explanation/intro-to/security.md
s-makin: Agreed, that sounds much better

explanation/intro-to/security.md
s-makin: I'm not sure we have a good single official source for this - except maybe the packaging guide. I've added a link to the packaging guide section on components, but can change it if there's a better one.

explanation/security/security_suggestions.md
s-makin: I think this is content the security team is planning to include in their new/expanded docs - once I have a link I can include it here.

explanation/security/security_suggestions.md
s-makin: Sounds good, I can add it :)

explanation/security/security_suggestions.md
s-makin: I've added a little paragraph about it, but since the pages I highlighted in my comment are all in the cryptography section (mentioned later) I'm going to just leave it at that. I believe that the security team intends to have some content about TLS/SSL so again, once that exists, I'll link to it from here.

s-makin:
> Provided some in-line comments to further guide you. You said it isn't yet fully ready, so I'm neither approving nor rejecting at this stage.

Thanks for the initial comments, they were really helpful. At the time I'd only managed to write up about 3/4 of the contextual text I wanted to include, but it's now fully included and I've incorporated your feedback as well. I'd say this is now ready to be reviewed as normal.

explanation/security/security_suggestions.md
cpaelzer: yeah since it is "external" I'm ok to wait for that to exist

cpaelzer (approved):
Hey, I like it.
Thanks for adding AD and ack to add the ssh hardening when it is in official docs.
What you have improves te landing and suggestion elements of security quite nicely.
And as intended we can add/link more

The only concern I have is that diataxis made these things appear in so many places.
The pages are good, but e.g. the fold-out in "Explanation -> security -> suggestions" is so much smaller than e.g. the list in "Advanced Access Control", "Security of communications" , ...
Normally you'd think an outline nav to match what is on the page, but this is the diataxis grouping which has pro's and con's here showing a Con - not a fault or problem with this PR.

Despite this slight navigation loss it provides nice "getting started" pages for the topic, which was the intent.

I'm +1 and we can land this, further polishing as requests or insights come up are for the future.

PR #97, opened 2024-12-14
./easyrsa gen-dh command generates dh.pem, not dh2048.pem

yukihane:


cpaelzer:
Hi and thanks @yukihane ,
I verified if it might have been a different name in different releases, but at least jammy, noble and plucky which I tried all behave exactly the same and indeed generate dh.pem, not dh2048.pem.

PR #99, opened 2024-12-17
Revert "Change install-nvidia-drivers.md to have correct --gpgpu posi…

MitchellAugustin:
…tion order"

This reverts commit 4a2598326ad789b188523c5b7954e8f758afde90.

The 'this is gpgpu mode' print upon which this commit was based is inaccurate - in reality, the original ordering is the correct ordering, as can be shown by running
`sudo ubuntu-drivers list --gpgpu --recommended`
and seeing the -server driver recommended, whereas with 
`sudo ubuntu-drivers --gpgpu list --recommended`
the UDA driver is recommended

### DIFF
--- a/how-to/graphics/install-nvidia-drivers.md
+++ b/how-to/graphics/install-nvidia-drivers.md
@@ -38,7 +38,7 @@ sudo ubuntu-drivers list
 or, for servers:
 
 ```bash
-sudo ubuntu-drivers --gpgpu list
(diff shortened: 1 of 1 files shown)


cpaelzer:
I wonder which steps we should do for verifying doc contributions about things that need special HW other than trusting you and other.

Thanks for helping to update this. You did not lose karma points as you fixed your mistake yourself :-)

PR #101, opened 2024-12-20
Change console highlight syntax to bash

### LINKED ISSUE #40: Formatting: code blocks
Many pages have code blocks declared as "console", but the resulting rendered documentation has too much syntax highlighting which can make it harder to parse when reading.

It would be much better if these "console" code blocks were changed to "bash" for consistency with the rest of the docs. If there are any code blocks that have no declared language, it would be good to add one (bash, text for output, yaml, etc).

lvoytek:
Closes #40

This update changes all the "console" syntax highlights to "bash"
For example, code blocks tend to go from

image

to

image

LMK if you think this is better. Also as a side note, it would probably be a good idea to clean up code blocks such that all commands consistently include or remove the `$` at the beginning. That should also help clean up the syntax highlighting a bit.

### DIFF
--- a/explanation/crypto/gnutls.md
+++ b/explanation/crypto/gnutls.md
@@ -83,7 +83,7 @@ A priority string can start with a single initial keyword, and then add or remov
 
 To see the resulting list of ciphers and algorithms from a priority string, one can use the `gnutls-cli` command-line tool. For example, to list all the ciphers and algorithms allowed with the priority string `SECURE256`:
 
-```console
+```bash
 $ gnutls-cli --list --priority SECURE256
 Cipher suites for SECURE256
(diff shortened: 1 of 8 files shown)


panlinux:
This change looks good, thanks.
Regarding the prompt in code blocks, i.e., `$ command <args>`, I very much dislike the pattern where we first show the command (without the prompt), and then the result. I understand it helps with copy/paste, but I think it just looks bad. I would very much prefer to have a different style for what you type, and what you get in response, which is how all technical books I have read in my life are. Whatever the choice, however, I agree we must be consistent throughout. This is worth a separate issue.

panlinux (approved):
+1, it looks better

panlinux:
I also filed https://github.com/canonical/ubuntu-server-documentation/issues/102 for command/output consistency.

PR #113, opened 2025-01-11
Fix AppArmor documentation for total enable/disable of AppArmor (Fixes #105)

### LINKED ISSUE #105: doc: Inaccurate/Incomplete "Unload/Disable AppArmor" documentation?
The AppArmor documentation stated with the SystemD option to disable AppArmor and unload/disable the Kernel module doesn't appear to work - per user statements here: https://askubuntu.com/questions/1536716/why-cant-i-disable-apparmor-on-24-04-1?noredirect=1#comment2703457_1536716

However, lifting from the original documentation on AppArmor on the Wiki (ancient! - https://help.ubuntu.com/community/AppArmor#Disable_AppArmor_framework) it includes how to disable AppArmor in GRUB's defaults. According to the user in the Ask Ubuntu statement, the SystemD approach did not disable the AppArmor kernel module nor unload it even on reboot with the service disabled.

It took revisions to /etc/default/grub per the original AppArmor wiki page to properly disable the AppArmor kernel modules.

This may be a bug in the SystemD task, or something that is overlooked in the Server documentation, but needs reviewed for accuracy - if it's not accurate here, then the documentation needs updated; if it's a bug in the SystemD tasks then Kernel and Security teams need to be aware of the broken service load/unload issues.

---
Document: how-to/security/apparmor.md

teward:
AppArmor disabling/enabling is no longer just a single `systemctl` command to start or stop the service.  The AppArmor module and service are baked into the Kernel.

As such, it is no longer a single SystemD command to stop the service, as you also have to edit the GRUB and boot configurations accordingly to disable this in the Kernel.

Refer to #105 where I brought up this concern, and it was confirmed this is inaccurate.  This PR provides proposed language / sections to address that issue.

**This branch and its data are allowed to be edited by maintainers per the Pull Request**.

### DIFF
--- a/how-to/security/apparmor.md
+++ b/how-to/security/apparmor.md
@@ -230,29 +230,21 @@ To disable AppArmor, you must do the following:
     sudo systemctl disable apparmor.service
     ```
 
-* Edit `/etc/default/grub` and add the following to the GRUB configuration to disable the AppArmor kernel module from loading at boot time:
-
-    ```
-    apparmor=0
-    security=""
-    ```
+* Edit `/etc/default/grub` and add `apparmor=0` to `GRUB_CMDLINE_LINUX` in `/etc/default/grub`
 
 * Run `sudo update-grub` to refresh the boot configuration
 * Reboot the system.
 
 ### Re-enable AppArmor
 
-* Remove the following lines from `/etc/default/grub` or comment them out by putting a `#` at the beginning of the lines:
-    ```
-    apparmor=0
-    security=""
-    ```
+* Remove the `apparmor=0` item from `GRUB_CMDLINE_LINUX` in `/etc/default/grub`.
 
 * Re-enable the `apparmor.service` SystemD unit
 
     ```
     sudo systemctl enable apparmor.service
     ```
+    
 * Run `sudo update-grub` to update your system boot configuration.
 * Reboot your system.
 
--- a/how-to/security/apparmor.md
+++ b/how-to/security/apparmor.md
@@ -76,20 +76,6 @@ sudo rm /etc/apparmor.d/disable/profile.name
 cat /etc/apparmor.d/profile.name | sudo apparmor_parser -a
 ```
 
-AppArmor can be disabled, and the kernel module unloaded, by entering the following:
-
-```bash
-sudo systemctl stop apparmor.service
-sudo systemctl disable apparmor.service
-```
-
-To re-enable AppArmor, enter:
-
-```bash
-sudo systemctl enable apparmor.service
-sudo systemctl start apparmor.service
-```
-
 > **Note**:
 > Replace `profile.name` with the name of the profile you want to manipulate. Also, replace `/path/to/bin/` with the actual executable file path. For example, for the `ping` command use `/bin/ping`.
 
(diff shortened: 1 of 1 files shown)


how-to/security/apparmor.md
```diff
@@ -225,6 +211,43 @@ Profiles are meant to provide security and so can't be too permissive. But often
...
   * The files can be found in `/etc/apparmor.d/local/` and exist for the packages that are known to sometimes need slight tweaks for special setups.
 
+
+## Disabling or Re-enabling AppArmor
+
+Starting with Ubuntu 24.04 and later, the AppArmor services are baked into the Ubuntu Kernel, and require special steps to fully disable or re-enable.
+
+### Disable AppArmor
+
+> WARNING! Disabling AppArmor reduces the security of your system! You should only disable apparmor if you understand the security implications
```
cpaelzer, line 221: I like adding a warning banner - thanks

how-to/security/apparmor.md
```diff
@@ -225,6 +211,43 @@ Profiles are meant to provide security and so can't be too permissive. But often
   * To mitigate the drawbacks of above approaches, **local includes** were introduced, adding the ability to write arbitrary rules that not run into issues during upgrades that modify the packaged rule.
   * The files can be found in `/etc/apparmor.d/local/` and exist for the packages that are known to sometimes need slight tweaks for special setups.
 
+
+## Disabling or Re-enabling AppArmor
+
+Starting with Ubuntu 24.04 and later, the AppArmor services are baked into the Ubuntu Kernel, and require special steps to fully disable or re-enable.
```
cpaelzer, line 217: You and me know what you mean, but for others it might be "which are the *special* steps".
Maybe:

```suggestion
Starting with Ubuntu 24.04 and later, the AppArmor services are baked into the Ubuntu Kernel. While before you disable it by not loading the service, it now requires setting a module parameter on the kernel command line to fully disable or re-enable.
```

how-to/security/apparmor.md
```diff
@@ -225,6 +211,43 @@ Profiles are meant to provide security and so can't be too permissive. But often
...
+
+### Disable AppArmor
+
+> WARNING! Disabling AppArmor reduces the security of your system! You should only disable apparmor if you understand the security implications 
+> of disabling the service!
+
+To disable AppArmor, you must do the following:
+
+* Stop and disable the `apparmor.service` in SystemD to prevent AppArmor profiles from being loaded:
```
cpaelzer, line 226: Due to AssertPathIsReadWrite=/sys/kernel/security/apparmor/.load
We actually could "only" document the module argument now.
That will avoid the service to be enabled/disables - less steps, less mistakes.

Also re-enabling it will be only one step again.
WDYT - should we just remove all about apparmor.service?

cpaelzer: Because disabling it with a kernel parameter works in the old style as well.
There you could provide the parameter via a conf file, but since disabling security is nothing you'd want to make too trivial just and only documenting the kernel parameter might be better readable and works everywhere.

how-to/security/apparmor.md
```diff
@@ -225,6 +211,43 @@ Profiles are meant to provide security and so can't be too permissive. But often
...
+    sudo systemctl stop apparmor.service
+    sudo systemctl disable apparmor.service
+    ```
+
+* Edit `/etc/default/grub` and add `apparmor=0` to `GRUB_CMDLINE_LINUX` in `/etc/default/grub`
+
+* Run `sudo update-grub` to refresh the boot configuration
+* Reboot the system.
+
+### Re-enable AppArmor
```
cpaelzer, line 238: maybe we should mention how an enabled apparmor looks like
How about adding as example:

```
$ sudo aa-status
apparmor module is loaded.
119 profiles are loaded.
24 profiles are in enforce mode.
   /usr/bin/man
...
```

how-to/security/apparmor.md
```diff
@@ -225,6 +211,43 @@ Profiles are meant to provide security and so can't be too permissive. But often
...
+
+    ```
+    sudo systemctl stop apparmor.service
+    sudo systemctl disable apparmor.service
+    ```
+
+* Edit `/etc/default/grub` and add `apparmor=0` to `GRUB_CMDLINE_LINUX` in `/etc/default/grub`
+
+* Run `sudo update-grub` to refresh the boot configuration
+* Reboot the system.
```
cpaelzer, line 236: We should add a bit how one can see if it worked.
I got the output of aa-status and the service in that case.
I think it could be valuable to all to be sure what they did worked?

```
$ sudo aa-status
apparmor module is loaded.
apparmor filesystem is not mounted.

$ systemctl status apparmor
○ apparmor.service - Load AppArmor profiles
     Loaded: loaded (/usr/lib/systemd/system/apparmor.service; enabled; preset: enabled)
     Active: inactive (dead)
  Condition: start condition unmet at Tue 2025-01-14 08:45:04 UTC; 1min 20s ago
             └─ ConditionSecurity=apparmor was not met
       Docs: man:apparmor(7)
             https://gitlab.com/apparmor/apparmor/wikis/home/

Jan 14 08:45:04 n systemd[1]: apparmor.service - Load AppArmor profiles was skipped because of an unmet condition check (ConditionSecurity=apparmor).
```

cpaelzer (changes requested):
I've left a suggestion and some questions.
This is already good (no surprise as you know what you do :-) ) but it can be even better while touching it. Please have a look.

And once you are done adapting to where you agree, please consider
- squashing the commits together
- adding a `Fixes: #105` to the commit

how-to/security/apparmor.md
teward: Would it work to have a section titled "Checking AppArmor status" dedicated to this?

how-to/security/apparmor.md
teward: see https://github.com/canonical/ubuntu-server-documentation/pull/113#discussion_r1915515239

how-to/security/apparmor.md
teward: If we don't need `apparmor.service` to be disabled manually then we can remove that bit and just use the unloading of the module.  Same for re-enabling, because then the service works as intended if the module is present or not.

teward: If we don't need `apparmor.service` to be disabled manually then we can remove that bit and just use the unloading of the module.  Same for re-enabling, because then the service works as intended if the module is present or not.

teward:
(When merging please squash all commits into one)

cpaelzer (commented):
Thanks @teward it now has all I asked for and works for older and newer release just as fine.
Finally the addition of the examples helps to ensure the user has done it right.

Merging ...

PR #130, opened 2025-01-31
lxc-containers: remove page

Sophie-Pages:
- Remove LXC containers page.
- Remove broken links.
- Redirect to official LXC documentation.
- Because Rediraffe doesn't support external urls (see [open issue](https://github.com/wpilibsuite/sphinxext-rediraffe/issues/41)), I couldn't add https://linuxcontainers.org/lxc/documentation/ in redirects.txt. I put the redirection in custom_conf.py's redirects dict.


s-makin (approved):
This looks good to me, thanks @Sophie-Pages!

I'm not 100% sure if the redirect will work since the previews on RTD can be a bit awkward, I will have to check after it's merged and maybe set up a manual redirect in RTD itself (hopefully it will work just fine :crossed_fingers:)

s-makin:
> This looks good to me, thanks @Sophie-Pages!
> 
> I'm not 100% sure if the redirect will work since the previews on RTD can be a bit awkward, I will have to check after it's merged and maybe set up a manual redirect in RTD itself (hopefully it will work just fine 🤞)

Just as a follow up, I saw from the traffic to the project that the redirect set up via conf.py didn't work. I've now set one up instead in the RTD admin panel (which is now working). I'm going to update the contributing guide to reflect this.

Sophie-Pages:
Oh that's unfortunate! After running `make clean` and `make run` the redirection worked for me so I was hopeful. Thank you @s-makin for the follow up!

s-makin:
> Oh that's unfortunate! After running `make clean` and `make run` the redirection worked for me so I was hopeful. Thank you @s-makin for the follow up!

No problem :) after a little bit of fiddling around I discovered that the sphinx-reredirect extension (which is what's backing the redirects in the custom_conf.py file) expects the full file path, so after I updated the redirect to this:
```
"how-to/containers/lxc-containers": "https://linuxcontainers.org/lxc/documentation/"
```
Then it worked on the live docs :)

PR #131, opened 2025-02-03
Improve the intro to ebf page

network-charles:
- I added a language for the texts wrapped into a [CodeFence](https://myst-parser.readthedocs.io/en/v0.13.6/using/syntax.html#commonmark-tokens).
- In Line 146 since "files" is a plural, I changed a singular "is" to a plural "are".
- In Line 193, I changed "across all of the system" to "across all the system" to make the sentence concise.

Relevant to https://github.com/canonical/open-documentation-academy/issues/162

### DIFF
--- a/explanation/intro-to/ebpf.md
+++ b/explanation/intro-to/ebpf.md
@@ -60,7 +60,7 @@ Ubuntu](https://discourse.ubuntu.com/t/spec-include-performance-tooling-in-ubunt
 different functionalities. Apart from the `bpftrace` tool itself, you can fetch
 a comprehensive list of these tools with the following command:
 
-```
+```bash
 $ dpkg -L bpftrace bpfcc-tools | grep -E '/s?bin/.*$' | xargs -n1 basename
 ```
 
@@ -83,7 +83,7 @@ output for `bpftrace` tools in
 
 For instance,
 
-```
+```bash
 # bashreadline.bt
 ```
 
@@ -111,7 +111,7 @@ For that you'd run `execsnoop-bpfcc` with the following arguments:
 * `-Uu root` - to reduce the noisy output only to things done in root context (like here the package install)
 * `-T` - to get time info along the log
 
-```
+```bash
 # In one console run:
 $ sudo execsnoop-bpfcc -Uu root -T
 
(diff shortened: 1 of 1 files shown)


explanation/intro-to/ebpf.md
```diff
@@ -143,7 +143,7 @@ TIME     UID   PCOMM            PID     PPID    RET ARGS
 Let us look at another practical application of eBPF. This time meant to show
 another use-case, but also evolve it into more by modifying it.
 
-### Which files is my QEMU loading?
+### Which files are my QEMU loading?
```
s-makin, line 146: This change will need to be reverted - since QEMU is the subject of the verb, and it's singular, the original is correct :)

explanation/intro-to/ebpf.md
```diff
@@ -190,8 +190,7 @@ $ lxc launch ubuntu-daily:n n-vm-test --ephemeral --vm
 
 Of course the QEMU process opens plenty of things: shared libraries, config
 files, entries in `/{sys,dev,proc}`, and much more. But here we can see them all
-as they happen across all of the system.
-
+as they happen across all the system.
```
s-makin, line 193: I think a better change here would be to avoid the double use of "all" by changing the sentence to:
"But here we can see them all as they happen across the whole system"
What do you think?

s-makin (commented):
Thanks for taking the time to help make the Server docs better :) greater consistency in how the languages are defined in code blocks was a [previously filed](https://github.com/canonical/ubuntu-server-documentation/issues/40) (and closed) issue, so thanks for finding these additional ones that were missed.

I do have two other comments, but once those are addressed I will be happy to merge these changes.

explanation/intro-to/ebpf.md
network-charles: Nice, I like that.

explanation/intro-to/ebpf.md
network-charles: Ah, I see now. Alright

s-makin (approved):
This looks good to me :) thanks for making the suggested changes, and for continuing to help us make the Server docs better

PR #134, opened 2025-02-07
Update glossary b

RomainDusi:


reference/glossary.rst
```diff
@@ -304,220 +304,143 @@ Glossary
         *Work in Progress*
 
     DAC
-        *Work in Progress*
+        Discretionary access control
+        A form of access control where the owner of a resource can grant/revoke permissions to other users.
```
s-makin, line 308:
```suggestion
        Discretionary access control

        A form of access control where the owner of a resource can grant/revoke permissions to other users.
```

s-makin: In the rendered documentation page, the line breaks you've put here are not being respected and all the text is running together. I think you might need to try putting empty lines between any lines that you want to separate on the page

s-makin: An alternative could be to have any expanded acronym formatted a particular way (e.g. in bold face), and then a colon to separate the description? E.g.
```
**Discretionary Access Control**: A form of...
```
Since we don't have a standard formatting yet, I'll let you decide :)

explanation/crypto/bind-9-dnssec-cryptography-selection.md
```diff
@@ -1,7 +1,7 @@
 (bind-9-dnssec-cryptography-selection)=
 # BIND 9 DNSSEC cryptography selection
 
-Domain Name System Security Extensions (DNSSEC), which provides a set of security features to DNS, is a broad topic. In this article, we will briefly show DNSSEC validation happening on a `bind9` DNS server, and then introduce the topic of how we can disable certain cryptographic algorithms from being used in this validation.
+Domain Name System Security Extensions (DNSSEC), which provides a set of security features to [DNS](https://documentation.ubuntu.com/server/reference/glossary/#term-DNS), is a broad topic. In this article, we will briefly show DNSSEC validation happening on a `bind9` DNS server, and then introduce the topic of how we can disable certain cryptographic algorithms from being used in this validation.
```
s-makin, line 4: Because we're using MyST markdown there's a specific role we can use to define glossary terms. Once the hoverxref functionality has been activated (this is in progress), using those roles will allow us to have Wikipedia style hover-over boxes that will pull the term definitions from the glossary and show them on the page (so people won't need to click to follow the link, they can just hover).

To avoid any need for refactoring after that's activated, please could you change the formatting of the links from the standard markdown [DNS](hardcoded/url/to/term) with:
```
{term}`DNS`
```
The matching/linking of terms to the glossary is very conveniently handled by Sphinx when the docs are rendered so we don't need to link any of the terms specifically :)

s-makin (commented):
Thank you so much for kicking us off with this! This looks like it was a lot of work, but I really love the definitions you've added to the glossary list, they're clear but concise :ok_hand: 

There is a bit of weirdness in the glossary page is rendered in the preview, in that the newlines you put haven't been respected. I've got some suggestions to fix that (see below).

The work you've done on linking to the correct instances of jargon terms on each page is exactly what I was looking for, thank you! I do just have one little request on the formatting of the links. Sphinx is smart enough to link up the terms with their glossary definitions (as long as the term matches what's in the glossary), so we don't need to hardcode the links. If you'd like me to add these as review suggestions (to make committing them easier) I'm happy to do so, please let me know :)

tutorial/managing-software.rst
```diff
@@ -651,7 +651,7 @@ or later, we'll want to customise the package so that it better fits our own
 purposes.
 
 Before we try to customise the package, we should probably look at what files
-are included in it. We can check this using ``dpkg``, which is the
+are included in it. We can check this using `dpkg <https://documentation.ubuntu.com/server/reference/glossary/#term-dpkg>`_, which is the
```
s-makin, line 654:
```suggestion
are included in it. We can check this using ``dpkg``, which is the
```

explanation/intro-to/openldap.md
```diff
@@ -8,15 +8,15 @@ that it's understood as such.
 
 ## Key concepts and terms
 
-- A **directory** is a tree of data **entries** that is hierarchical in nature; it is called the Directory Information Tree (DIT).
+- A **directory** is a tree of data **entries** that is hierarchical in nature; it is called the [Directory Information Tree (DIT)](https://documentation.ubuntu.com/server/reference/glossary/#term-DIT).
```
s-makin, line 11:
```suggestion
- A **directory** is a tree of data **entries** that is hierarchical in nature; it is called the Directory Information Tree ({term}`DIT`).
```

explanation/software/third-party-repository-usage.md
```diff
@@ -14,7 +14,7 @@ Although we don’t recommend using third party software, we know that users som
 
 When using any software that you have not audited yourself, you must implicitly trust the publisher of that software with your data. However, with third party APT repositories, there are additional implications of this that are less obvious.
 
-Unlike more modern packaging systems, APT repositories run code that is not sandboxed. When using software from more than one publisher, such as from your distribution as well as a third party, APT and dpkg provide no security boundary between them.
+Unlike more modern packaging systems, APT repositories run code that is not sandboxed. When using software from more than one publisher, such as from your distribution as well as a third party, APT and [dpkg](https://documentation.ubuntu.com/server/reference/glossary/#term-dpkg) provide no security boundary between them.
```
s-makin, line 17:
```suggestion
Unlike more modern packaging systems, APT repositories run code that is not sandboxed. When using software from more than one publisher, such as from your distribution as well as a third party, APT and {term}`dpkg` provide no security boundary between them.
```

s-makin (approved):
Really nice work, thanks @RomainDusi :) I took the liberty of fixing up one or two links that weren't linking correctly in the rendered docs, and now it's all working as expected. Really love the definitions you added, and thanks for the additional tidying up you did as well. All LGTM :)

PR #135, opened 2025-02-07
mention about smbpasswd

cezary-z:
Make sure the user is added to the Samba user database

### DIFF
--- a/how-to/samba/share-access-controls.md
+++ b/how-to/samba/share-access-controls.md
@@ -34,6 +34,12 @@ After editing `/etc/samba/smb.conf`, reload Samba for the changes to take effect
 ```bash
 sudo smbcontrol smbd reload-config
 ```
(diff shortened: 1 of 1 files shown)


panlinux:
Thanks for this. Indeed, I searched the docs for `smbpasswd` and there is no mention of it in the context of a non-domain controller, i.e., a simple file server. In particular, there is no mention of the fact that you will need to have the same user (`melissa`) exist in linux (in `/etc/passwd` and `/etc/shadow`, for example), **AND** in samba, which is where the `smbpasswd` command comes in (with `-a` very likely, to create the user in the samba DB).

Do you think you could expand on that a bit for this change? Like a new section near the top, quickly explaining that in the isolated server scenario, we need to create the user both in the linux world, and in the samba world?

panlinux:
Thank you for having created this PR. I ended up expanding a bit on the concept of Samba users in PR #290 and will close this one in favor of that.

PR #136, opened 2025-02-08
Added new how-to page in the ldap section for setting up ldap with si…

davidstraight:
…mple pass-through auth using saslauthd and kerberos, also added entry in the table of contents for the ldap section

### DIFF
--- a/how-to/openldap.rst
+++ b/how-to/openldap.rst
@@ -9,6 +9,7 @@ OpenLDAP
 * :ref:`Simple LDAP user and group management <ldap-users-and-groups>`
 * :ref:`OpenLDAP and Transport Layer Security (TLS) <ldap-and-tls>`
 * :ref:`Backup and restore OpenLDAP <ldap-backup-and-restore>`
+* :ref:`Set up OpenLDAP with plain auth using saslauthd and Kerberos <ldap-saslauthd-kerberos>`
 
 .. toctree::
     :hidden:
@@ -19,3 +20,5 @@ OpenLDAP
     LDAP users and groups <openldap/users-and-groups>
     LDAP and TLS <openldap/ldap-and-tls>
     Backup and restore <openldap/backup-and-restore>
+    LDAP plain auth with Kerberos <openldap/ldap-saslauthd-kerberos>
+
--- a/how-to/openldap/ldap-saslauthd-kerberos.md
+++ b/how-to/openldap/ldap-saslauthd-kerberos.md
@@ -1,0 +1,282 @@
+(ldap-saslauthd-kerberos)=
+# How to configure OpenLDAP with pass-through SASL authentication using Kerberos
+There exists applications that can only use "simple" username and password [Lightweight Directory Access Protocol](https://ldap.com/) (LDAP) authentication. Pass-though password authentication is a solution to this problem.
+
+## Before you begin
+It is assumed you are starting with a working OpenLDAP server, with a hostname of `ldap-server.example.com`. If not, follow this guide {ref}`Install and configure OpenLDAP<install-openldap>` to set it up. It is also assumed that the `EXAMPLE.COM` realm is set up, and the Kerberos client tools (krb5-user) are installed on the ldap server. You will need to create an ubuntu principal. See {ref}`How to install a Kerberos server <install-a-kerberos-server>`. You should also know how to create service principals. See {ref}`How to configure Kerberos service principals <configure-service-principals>`.
+All the followint configuration will be on `ldap-server.example.com`.
+> **Note**:
+> This process is not the same as using [Generic Security Services Application Programming Interface](https://www.openldap.org/doc/admin26/sasl.html#GSSAPI) (GSSAPI) to log into the LDAP server.
+> Rather it is using [simple authentication](https://www.openldap.org/doc/admin26/security.html#%22simple%22%20method) with the OpenLDAP server so this should be over a [Transport Layer Security](https://datatracker.ietf.org/wg/tls/documents/) (TLS) connection.
+> This process also does not require the Kerberos backend to be OpenLDAP server, the default Keberberos backend if sufficient.
+> The test user we will be using is `ubuntu@EXAMPLE.COM` which must exist in the Kerberos database
+
+
+## Package installation
+Install saslauthd
+
+```bash
+sudo apt install sasl2-bin
+```
+## Saslauthd service principal
+It is assumed you have an admin principal set up in the realm. We will use `ubuntu/admin@EXAMPLE.COM`.
+Next a principal who will be querying Kerberos is needed. The default for saslauthd is "host"/"hostname.domain"@"REALM". For the basic case, it is important that the hostname and domain be what nslookup on your IP number says it is.
+
+## Check the hostname
+Get the hostname from the server
+```bash
+hostname
+```
+Which should give you the hostname of:
+```text
+ldap-server
+```
+Next check what nslookup gives as the full canonical name
+```bash
+nslookup ldap-server
+```
+Which should look something like this.
+```text
+Non-authoritative answer:
+Name:   ldap-server.example.com
+Address: (your IP)
+```
+Finally, check the hostname and domain using a reverse lookup.
+```bash
+nslookup (your IP)
+```
+Check the name in the reply
+```text
+(PI ruoy).in-addr.arpa      name = ldap-server.example.com. (PI ruoy) is (your IP) in reverse
+```
+If the result is the same as your host's canonical name them all is well. If the domain is missing, the [Fully Qualified Domain Name](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) (FQDN) can be entered in the `/etc/hosts` file.
+```bash
+sudo vi /etc/hosts
+```
+Add the FQDN before the short hostname.
+```text
+(your IP) ldap-server.example.com ldap-server
+```
+## Create the saslauthd principal
+
+The hostname portion of the principal must be in lowercase. You won't be creating a password for this principal so you will need to ask the kadmin tool to store the key in an external file on the LDAP server. The default location is `/etc/krb5.keytab` so you need to run the kadmin command as sudo.
+
+```bash
+sudo kadmin -p ubuntu/admin
+Authenticating as principal ubuntu/admin with password.
+Password for ubuntu/admin@EXAMPLE.COM
+kadmin: addprinc -randkey host/ldap-server.example.com
+No policy specified for host/ldap-server.example.com@EXAMPLE.COM; defaulting to no policy
+Principal "host/ldap-server.example.com@EXAMPLE.COM" created.
+kadmin: ktadd host/ldap-server.example.com
+Entry for principal host/ldap-server.example.com with kvno 2, encryption type aes256-cts-hmac-sha1-96 added to keytab FILE:/etc/krb5.keytab.
+Entry for principal host/ldap-server.example.com with kvno 2, encryption type aes128-cts-hmac-sha1-96 added to keytab FILE:/etc/krb5.keytab.
+kadmin: q
+```
+To check our principals enter:
+```bash
+sudo klist -k
+```
+You should see the following.
+```text
+Keytab name: FILE:/etc/krb5.keytab
+KVNO Principal
+---- --------------------------------------------------------------------------
+   2 host/ldap-server.example.com@EXAMPLE.COM
+   2 host/ldap-server.example.com@EXAMPLE.COM
+```
+## Configure saslauthd
+Next look at the default saslauthd file located in `/etc/default/saslauthd`
+```bash
+sudo cat /etc/default/saslauthd
+```
+Here is the default config.
+```text
+#
+# Settings for saslauthd daemon
+# Please read /usr/share/doc/sasl2-bin/README.Debian for details.
+#
+
+# Description of this saslauthd instance. Recommended.
+# (suggestion: SASL Authentication Daemon)
+DESC="SASL Authentication Daemon"
+
+# Short name of this saslauthd instance. Strongly recommended.
+# (suggestion: saslauthd)
+NAME="saslauthd"
+
+# Which authentication mechanisms should saslauthd use? (default: pam)
+#
+# Available options in this Debian package:
+# getpwent  -- use the getpwent() library function
+# kerberos5 -- use Kerberos 5
+# pam       -- use PAM
+# rimap     -- use a remote IMAP server
+# shadow    -- use the local shadow password file
+# sasldb    -- use the local sasldb database file
+# ldap      -- use LDAP (configuration is in /etc/saslauthd.conf)
+#
+# Only one option may be used at a time. See the saslauthd man page
+# for more information.
+#
+# Example: MECHANISMS="pam"
+MECHANISMS="pam"
+
+# Additional options for this mechanism. (default: none)
+# See the saslauthd man page for information about mech-specific options.
+MECH_OPTIONS=""
+
+# How many saslauthd processes should we run? (default: 5)
+# A value of 0 will fork a new process for each connection.
+THREADS=5
+
+# Other options (default: -c -m /var/run/saslauthd)
+# Note: You MUST specify the -m option or saslauthd won't run!
+#
+# WARNING: DO NOT SPECIFY THE -d OPTION.
+# The -d option will cause saslauthd to run in the foreground instead of as
+# a daemon. This will PREVENT YOUR SYSTEM FROM BOOTING PROPERLY. If you wish
+# to run saslauthd in debug mode, please run it by hand to be safe.
+#
+# See /usr/share/doc/sasl2-bin/README.Debian for Debian-specific information.
+# See the saslauthd man page and the output of 'saslauthd -h' for general
+# information about these options.
+#
+# Example for chroot Postfix users: "-c -m /var/spool/postfix/var/run/saslauthd"
+# Example for non-chroot Postfix users: "-c -m /var/run/saslauthd"
+#
+# To know if your Postfix is running chroot, check /etc/postfix/master.cf.
+# If it has the line "smtp inet n - y - - smtpd" or "smtp inet n - - - - smtpd"
+# then your Postfix is running in a chroot.
+# If it has the line "smtp inet n - n - - smtpd" then your Postfix is NOT
+# running in a chroot.
+OPTIONS="-c -m /var/run/saslauthd"
+```
+Two changes are needed to use Kerberos, update MECHANISMS to kerberos5 and add START=yes
+
+```bash
+sudo vi /etc/default/saslauthd
+```
+Make these changes.
+```text
+...
+# Which authentication mechanisms should saslauthd use? (default: pam)
+#
+# Available options in this Debian package:
+# getpwent  -- use the getpwent() library function
+# kerberos5 -- use Kerberos 5
+# pam       -- use PAM
+# rimap     -- use a remote IMAP server
+# shadow    -- use the local shadow password file
+# sasldb    -- use the local sasldb database file
+# ldap      -- use LDAP (configuration is in /etc/saslauthd.conf)
+#
+# Only one option may be used at a time. See the saslauthd man page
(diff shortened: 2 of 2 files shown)


how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
+(ldap-saslauthd-kerberos)=
+# How to configure OpenLDAP with pass-through SASL authentication using Kerberos
+There exists applications that can only use "simple" username and password [Lightweight Directory Access Protocol](https://ldap.com/) (LDAP) authentication. Pass-though password authentication is a solution to this problem.
+
+## Before you begin
+It is assumed you are starting with a working OpenLDAP server, with a hostname of `ldap-server.example.com`. If not, follow this guide {ref}`Install and configure OpenLDAP<install-openldap>` to set it up. It is also assumed that the `EXAMPLE.COM` realm is set up, and the Kerberos client tools (krb5-user) are installed on the ldap server. You will need to create an ubuntu principal. See {ref}`How to install a Kerberos server <install-a-kerberos-server>`. You should also know how to create service principals. See {ref}`How to configure Kerberos service principals <configure-service-principals>`.
+All the followint configuration will be on `ldap-server.example.com`.
```
panlinux, line 7: Typo: `following`

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+Next check what nslookup gives as the full canonical name
+```bash
+nslookup ldap-server
+```
+Which should look something like this.
+```text
+Non-authoritative answer:
+Name:   ldap-server.example.com
+Address: (your IP)
+```
```
panlinux, line 43: I suggest replacing the above with a call to `hostname -f`, that will use the same mechanism (append search domain to `ldap-server`, and query DNS).

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+Finally, check the hostname and domain using a reverse lookup.
+```bash
+nslookup (your IP)
+```
+Check the name in the reply
+```text
+(PI ruoy).in-addr.arpa      name = ldap-server.example.com. (PI ruoy) is (your IP) in reverse
```
panlinux, line 50: typo: `your IP`

how-to/openldap.rst
```diff
@@ -9,6 +9,7 @@ OpenLDAP
 * :ref:`Simple LDAP user and group management <ldap-users-and-groups>`
 * :ref:`OpenLDAP and Transport Layer Security (TLS) <ldap-and-tls>`
 * :ref:`Backup and restore OpenLDAP <ldap-backup-and-restore>`
+* :ref:`Set up OpenLDAP with plain auth using saslauthd and Kerberos <ldap-saslauthd-kerberos>`
```
panlinux, line 12: I'm thinking we can rearrange this a little bit. First, a document showing what this pass-through authentication is, using saslauthd. Basically when/why would we want it:
- the app may not support the more advanced auth mechanisms that openldap is setup to use, like GSSAPI. The app can only do simple bind over TLS
- or, your openldap is setup to use some other user database for authentication instead of the `userPassword` attribute (or GSSAPI), like local `/etc/shadow` via PAM. And you want to avoid having to sync passwords all over the place.
In essence, we are delegating the openldap simple bind authentication to a separate service: `saslauthd`.

Then list what this service can support. Of note I think there are only 3:
- kerberos (which you are describing)
- PAM (whatever pam is configured to do: could even be kerberos again)
- sasldb (not sure how common that is)

The others are:
- shadow: use PAM instead
- LDAP (would be using another ldap server). Maybe useful, maybe not? Unsure, but if someone wants to contribute ir, that works
- rimap: remote IMAP server. Sounds very corner case and ancient, but if someone wants to contribute it, that also works

This is where we could add a diagram. Something like:
openldap-saslauthd

Then we can dive in and explain how to use saslauthd with kerberos. The other options (PAM and sasldb) would be a separate PR, separate documents. But in this way we have the structure ready for expanding on them.

@s-makin we would have a top-level entry like "OpenLDAP and passthrough authentication" where we would explain the above, and subdocuments for each mechanism, where "kerberos" would be the first one, and proposed in this PR.

What do you think, @davidstraight ?

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+All the followint configuration will be on `ldap-server.example.com`.
+> **Note**:
+> This process is not the same as using [Generic Security Services Application Programming Interface](https://www.openldap.org/doc/admin26/sasl.html#GSSAPI) (GSSAPI) to log into the LDAP server.
+> Rather it is using [simple authentication](https://www.openldap.org/doc/admin26/security.html#%22simple%22%20method) with the OpenLDAP server so this should be over a [Transport Layer Security](https://datatracker.ietf.org/wg/tls/documents/) (TLS) connection.
+> This process also does not require the Kerberos backend to be OpenLDAP server, the default Keberberos backend if sufficient.
```
panlinux, line 11: What do you mean by kerberos backend, and default kerberos backend?

how-to/openldap.rst
s-makin: @panlinux that sounds like a sensible ordering. Would you want that done as part of this PR or in a subsequent one?

how-to/openldap/ldap-saslauthd-kerberos.md
davidstraight: The default Kerberos back-end is db2. This is the database that holds all of Kerberos' principals. There is a how-to in the docs explaining how to change that back-end to use the LDAP server itself. I wanted to be clear that this second step was not necessary to use saslauthd.

how-to/openldap.rst
davidstraight: @s-makin, @panlinux I could put version 1 of  that together in this PR if that wouldn't be too much for one PR.

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+# Example for non-chroot Postfix users: "-c -m /var/run/saslauthd"
+#
+# To know if your Postfix is running chroot, check /etc/postfix/master.cf.
+# If it has the line "smtp inet n - y - - smtpd" or "smtp inet n - - - - smtpd"
+# then your Postfix is running in a chroot.
+# If it has the line "smtp inet n - n - - smtpd" then your Postfix is NOT
+# running in a chroot.
+OPTIONS="-c -m /var/run/saslauthd"
+```
+Two changes are needed to use Kerberos, update MECHANISMS to kerberos5 and add START=yes
```
panlinux, line 155: I don't think `START=yes` is needed. This is a systemd service, you need to enable it via `systemctl enable saslauthd.service` (as you describe later).

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+sudo vi /etc/hosts
+```
+Add the FQDN before the short hostname.
+```text
+(your IP) ldap-server.example.com ldap-server
+```
+## Create the saslauthd principal
+
+The hostname portion of the principal must be in lowercase. You won't be creating a password for this principal so you will need to ask the kadmin tool to store the key in an external file on the LDAP server. The default location is `/etc/krb5.keytab` so you need to run the kadmin command as sudo.
+
```
panlinux, line 63: The default for this saslauthd service principal is indeed `host/HOSTNAME`, but it can be overridden in `/etc/saslauthd.conf`. For example, this is how it would be set to `saslauthd`, if so desired:
```
krb5_verify_principal: saslauthd
```

panlinux: I mean, maybe add a note about that. I do concede that perhaps we don't want to show too many options, though. I just thought in this case maybe it made sense as the `host` principal is used for so many other things, like ssh access.
If we come across other options that could make sense to override, maybe a quick section about `/etc/saslauthd.conf` on its own would be better, instead of notes here and there.
I'm still going through the doc, so maybe there are other opportunities I haven't reached yet.

davidstraight: I did actually make a list of advanced options that you can set in the saslauthd.conf file but I was worried about making it too confusing. But if the saslauthd how-to is in its own file then there would be more room to flesh out some advanced settings.

davidstraight: Example config options are `krb5_keytab`, `krb5_verify_principal`, `krb5_servername`, `krb5_uppercase_realm`, and the complicated one: `krb5_XXXX_instance` where XXX is a service passed into saslauthd and the value is what will be appended as the Kerberos instance. (`krb5_imap_instance: imap` will result in principals as `name/imap@realm` even if the user was passed in as name@realm.

how-to/openldap/ldap-saslauthd-kerberos.md
davidstraight: 22.04 still uses the init.d script so I was trying to be universal. Do we assume that people are using 24.04?

panlinux: Good point, I was trying 24.04. If 22.04 (and earlier) absolutely need the `START` option, then in such cases we assume the latest LTS is being used, and add a "note" highlighting the difference in older LTSs. For now use a `> **Note**:` block in this position in the doc. If these differences start to pile up, then we can consider a "Older LTSs" section to group them all.

how-to/openldap/ldap-saslauthd-kerberos.md
panlinux: > but I was worried about making it too confusing

That is a valid concern. Let's see if there are more config options that could be worthwhile to elaborate on, and then decide if we are going to detail them or not. I'm still going through the doc.

how-to/openldap.rst
panlinux: It can be a different PR, as long as this one is adapted to fit into the new structure. Which may be easier if that structure is in place already, or part of it.

how-to/openldap/ldap-saslauthd-kerberos.md
panlinux: Ah, I see, that's quite an advanced use-case anyway. I wouldn't have jumped to that conclusion, even though there is such a guide in the docs already. I think you make it clear above that the guide assumes the normal installation of both OpenLDAP and Kerberos servers, with links even. I think mentioning the way more advanced kerberos-backed-by-openldap is unnecessary here.

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+Continue by enabling the saslauthd service.
+
+```bash
+sudo systemctl enable saslauthd
+```
+The final step for saslauthd is to start the saslauthd service.
+
+```bash
+sudo systemctl start saslauthd
+```
```
panlinux, line 192: This would be a good point in the how-to do test saslauthd. This can be done with `testsaslauthd`. For example:
```
ubuntu@ldap-server:~$ testsaslauthd -u ubuntu -p ubuntusecret
0: OK "Success."
```
And with the wrong kerberos password:
```
ubuntu@ldap-server:~$ testsaslauthd -u ubuntu -p ubuntusecretwrong
0: NO "authentication failed"
```

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+# requesting: ALL
+#
+...
+```
+If you see search results then it worked, confirm using ldapwhoami
+```bash
+ldapwhoami -D uid=ubuntu,ou=People,dc=example,dc=com -W -x
+Enter LDAP Password:
+dn:uid=ubuntu,ou=People,dc=example,dc=com
+```
```
panlinux, line 274: I suggest to just use `ldapwhoami`, as it's shorter and needs less options, like a base search in `ldapsearch`, and the output of `ldapsearch` could be much longer.

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+sudo systemctl enable saslauthd
+```
+The final step for saslauthd is to start the saslauthd service.
+
+```bash
+sudo systemctl start saslauthd
+```
+## Configure OpenLDAP
+
+The OpenLDAP service needs its own Kerberos principal.
```
panlinux, line 195: Are you sure this is needed? I mean, it's needed for openldap native gssapi authentication, but if this setup is just using saslauthd, then it's not needed.

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+kadmin: q
+```
+The permissions of the keytab file need to be secure. Change the group and permissions to allow the slapd process to read the file.
+```bash
+sudo chown root:openldap /etc/ldap/ldap.keytab
+sudo chmod 640 /etc/ldap/ldap.keytab
+```
+OpenLDAP needs to know where to look for the keytab file, what password check method to use, and where to find saslauthd socket. This can be done using the sasl config file for the slapd.
+```bash
+sudo vi usr/lib/sasl2/slapd.conf
```
panlinux, line 215: This should be /etc/ldap/sasl2/slapd.conf. Maybe /usr/lib/sasl2/slapd.conf also works, but config files should be in /etc if possible.

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+homeDirectory: /home/ubuntu
+
+dn: cn=ubuntu,ou=Groups,dc=example,dc=com
+objectClass: posixGroup
+cn: ubuntu
+gidNumber: 10050
+memberUid: ubuntu
+LDIF
+```
+
```
panlinux, line 251: Can you add a brief explanation/summary here? I think we should point out the path this authentication is taking, and connect all the dots.

First, the simple bind via ldap is looking for the userPassword. But this server is configured via /etc/ldap/sasl2/slapd.conf to use `saslauthd` as the `pwcheck_method`, and the `userPassword` attribute is giving the sasl identity to use. Or maybe the order is different: when `userPassword` has content of the type `{SASL}<id>`, that's when `/etc/ldap/sasl2/slapd.conf` and `pwcheck_method` kicks in, and `saslauthd` is invoked.

`saslauthd` is configured to use `kerberos5` as the verification mechanism, and that invokes all the kerberos bits.

how-to/openldap/ldap-saslauthd-kerberos.md
```diff
@@ -0,0 +1,282 @@
...
+```
+If you see search results then it worked, confirm using ldapwhoami
+```bash
+ldapwhoami -D uid=ubuntu,ou=People,dc=example,dc=com -W -x
+Enter LDAP Password:
+dn:uid=ubuntu,ou=People,dc=example,dc=com
+```
+These LDAP users can now be used with external applications that only support "simple" username and password authentication.
+
+## References
```
panlinux, line 277: I think it would help to have a troubleshooting section. At least show how to bump the `saslauthd` log level via the `-d` option, which should not be set in `/etc/default/saslauthd` according to the comments in that file. In other words, to troubleshoot this service, it should first be stopped, then manually invoked via the command line where we can safely add `-d`.

how-to/openldap.rst
davidstraight: Good plan

how-to/openldap/ldap-saslauthd-kerberos.md
davidstraight: I believe the /usr/lib/sasl2/slapd.conf is the default. I don't know how to tell sasl to look elsewhere. I wanted to keep things in a default layout.

davidstraight: But I will test it to see if the /etc/ldap/sasl2/slapd.conf version works too

panlinux: Sorry, I forgot to say, I did test with `/etc/ldap/sasl2/slapd.conf`, and it worked.

davidstraight: That is very good to know. Should I even mention the default location if `/etc/ldap/sasl2` works just as well and is prefered?

how-to/openldap/ldap-saslauthd-kerberos.md
davidstraight: Ironically I used that option to figure out why it wasn't working for me at the beginning.

how-to/openldap/ldap-saslauthd-kerberos.md
davidstraight: I will verify

davidstraight: I just confirmed that you are correct and this is not needed.

how-to/openldap/ldap-saslauthd-kerberos.md
davidstraight: I agree. I could also show the result of a failed password in the `/var/log/auth.log`.  For example ` ...ldap-server saslauthd[936]:                 : auth failure: [user=ubuntu] [service=ldap] [realm=EXAMPLE.COM] [mech=kerberos5] [reason=krb5_get_init_creds_password failed]
`

how-to/openldap.rst
davidstraight: @panlinux would these documents still be in the open ldap folder or as a sub folder?

panlinux: So we are looking at an "openldap passthrough authentication" introduction, then this one, and in the future others showing the setup for other saslauthd backends (pam specially).
Let's experiment:
a) flat, all under how-to/openldap/:
- openldap-saslauthd-authentication-intro
- openldap-saslauthd-kerberos
- future: openldap-saslauthd-pam
b) A new directory: how-to/openldap-saslauthd-authentication/:
- openldap-saslauthd-authentication-intro
- openldap-saslauthd-kerberos
- future: openldap-saslauthd-pam
c) Like (b), but a new subdirectory inside how-to/openldap/

I don't have a strong opinion, and it looks like we can switch between these somewhat easily during development. I think we have a slight pattern for (b), for example, if you look at the samba docs. We have how-to/samba, but also how-to/active-directory, which is a big subtopic for samba. @s-makin , what do you think?

how-to/openldap/ldap-saslauthd-kerberos.md
panlinux: I have an idea for an image here, which would also help to show which bits of the process are done in clear text (and vulnerable), and which ones are using encryption (safe). I'll come back to this comment once I draw something.

panlinux:
Hi, just a note saying that I will come back to this PR soon. Apologies for the lack of responses.

davidstraight:
Sounds good. No rush.

panlinux:
I got allowed time to work on these bigger PRs/reviews on this next devel cycle, and will start next week.

how-to/openldap/ldap-saslauthd-authentication-intro.md
```diff
@@ -0,0 +1,32 @@
...
+
+## What is pass-through authentication
+A complication occurs when the passwords storage mechanism is outside of OpenLDAP. How can this simple bind process be used without storing the password in
+the OpenLDAP database. One solution to this problem to use use the Simple Authentication and Security Layer (SASL) library to pass though this bind request to the mechanism where the password is
+actually stored. The application that does this is the saslauthd daemon. This daemon can be configured to use various other authentication systems and will then
+pass back the authentication status to OpenLDAP which in turn passes it back to the calling application.
+
+A graphic showing this process is as follows:
+
+
```
panlinux, line 19: openldap + saslauthd diagram

How about using this image?

panlinux:
I'm going to resolve the conflicts here, merge, and later push follow-up smaller fixes.

Issue #125, opened 2025-01-22
[Feedback]: Can't access docs locally after a successful build


network-charles:
### Page URL

https://github.com/canonical/ubuntu-server-documentation/blob/main/contributing/build-locally.rst

### Select an option

- [ ] I found what I was looking for
- [ ] I couldn't find what I was looking for
- [ ] I found the information, but it was incorrect
- [x] I found the information, but it was incomplete
- [ ] I found the information, but it was confusing
- [x] I encountered a technical issue (e.g., broken link, image not loading)

### Issue Description

Everything builds successfully, but it fails when I access the URL, `http://127.0.0.1:8000/`.
The page refuses to load.

lvoytek:
Thank you for the report! I tried to reproduce and am able to access the URL after running `make run`. When I do, I see:

```
build succeeded.

The HTML pages are in _build.
[sphinx-autobuild] Serving on http://127.0.0.1:8000
[sphinx-autobuild] Waiting to detect changes...
```

Does this show up for you too?

network-charles:
Yes

network-charles:
Interesting. I tried two methods:
1. An Ubuntu server running on AWS
2. An Ubuntu docker image

network-charles:
**The running docs inside my container**
```bash
root@c2baef928c3d:/# make run

(good) pam-motd/index.html --> reference/other-tools/pam-motd/index.html
build succeeded.

The HTML pages are in _build.
[sphinx-autobuild] Serving on http://127.0.0.1:8000
[sphinx-autobuild] Waiting to detect changes...
```

**My docker container. I exposed port 8000.**
```bash
docker ps

➜  ~ docker ps
CONTAINER ID   IMAGE     COMMAND                  CREATED         STATUS         PORTS                    NAMES
c2baef928c3d   ubuntu    "bash -c 'while true…"   5 minutes ago   Up 5 minutes   0.0.0.0:8000->8000/tcp   gallant_robinson
➜  ~ 
```

**Error from my Safari browser.**
<img width="554" alt="Image" src="https://github.com/user-attachments/assets/2cabbd85-f1d2-4cde-9bdf-bc5d7576fa67" />

I can't reach the docs from Safari, but I can from inside the container via a `curl` command.

network-charles:
[...]

PR #138, opened 2025-02-09
Fix accessing built docs locally on Ubuntu cloud VM or container

network-charles:
This PR fixes the inability to access the docs locally in an Ubuntu cloud VM or container.

Relevant to #125

### DIFF
--- a/Makefile.sp
+++ b/Makefile.sp
@@ -54,7 +54,7 @@ sp-install: $(VENVDIR)
 	command -v distro-info || (sudo apt-get update; sudo apt-get install --assume-yes distro-info)
 
 sp-run: sp-install
-	. $(VENV); sphinx-autobuild -b dirhtml "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS)
+	. $(VENV); sphinx-autobuild -b dirhtml "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) --host 0.0.0.0
 
 # Doesn't depend on $(BUILDDIR) to rebuild properly at every run.
 sp-html: sp-install


network-charles:
Hi @lvoytek, I’d appreciate your feedback on this PR. Thanks

s-makin (commented):
I'm not sure that this is a change that we ought to make to the default Makefile, but if the change resolved your issue it might be worth documenting in the contributor guide as a helpful tip for others using cloud VMs to run the docs. Is this something you would be willing to add?

network-charles:
Yes, I can add it to this [file](https://github.com/canonical/ubuntu-server-documentation/blob/main/contributing/build-locally.rst).

contributing/build-locally.rst
```diff
@@ -82,6 +82,14 @@ It will watch the folder, and whenever you save changes to a file, this URL
 will give update the preview to show your changes (or warn you if something has
 gone horribly wrong!).
 
+If you are building locally on an Ubuntu Cloud VM or a container, you may experience issues accessing the page from a browser. To 
+resolve this, include `--host 0.0.0.0` in the `sp-run` section of the `Makefile.sp` file.
```
s-makin, line 86:
```suggestion
resolve this, include ``--host 0.0.0.0`` in the ``sp-run`` section of the ``Makefile.sp`` file.
```

contributing/build-locally.rst
```diff
@@ -82,6 +82,14 @@ It will watch the folder, and whenever you save changes to a file, this URL
...
+
+.. code-block:: sp
```
s-makin, line 88:
```suggestion
.. code-block::
```

s-makin (commented):
Thanks for making these changes, @network-charles. I only have a couple of formatting nits, but I'll commit those and then merge.

network-charles:
Hi Sally, new changes have been made to an upstream repo about this issue. It’s makefile was updated.

Canonical/sphinx-docs-starter-pack#353

I had a similar issue in the Juju docs so I was instructed to fix it upstream first. I was told that each makefile uses it as a reference. If that’s correct for the Server docs, can we update it here too?

s-makin:
> Hi Sally, new changes have been made to an upstream repo about this issue. It’s makefile was updated.
> 
> [canonical/sphinx-docs-starter-pack#353](https://github.com/canonical/sphinx-docs-starter-pack/pull/353)
> 
> I had a similar issue in the Juju docs so I was instructed to fix it upstream first. I was told that each makefile uses it as a reference. If that’s correct for the Server docs, can we update it here too?

We're running on an older/more stable version of the starter pack, so at some point we will need to update to a newer one (once I have the time to sit and go through all the changes to find a stable version we want to pin to, I'm going to migrate us to the new extension). At that point, the change in the linked PR will be included by default.

However I have to ask, does this change actually solve the problem you were having? Is your default makefile not being continually overwritten by the repo's default values whenever you pull/update?

network-charles:
Ah, nice!

Yes, it solves it. Once I permanently add the export variable to my shell “~/.bashrc”, I don't need to edit the makefile any time I do a make run anymore.

s-makin:
> Ah, nice!
> 
> Yes, it solves it. Once I permanently add the export variable to my shell “~/.bashrc”, I don't need to edit the makefile any time I do a make run anymore.

In that case, if you have the time and you want to bring over this change to the server docs I think that'd be great. If you do decide to do that, you'd probably need to also update the change you made to the contributing docs in *this* PR. Just to make sure everything's connected in the right places :)

network-charles:
Great, I will.

Issue #65, opened 2024-10-09
Improvement to LDAP Command Examples in Ubuntu Documentation


danruggi:
*Please describe the question or issue you're facing with "OpenLDAP replication - Ubuntu Server documentation".*

Hi,

While following the guide, I found an area for improvement regarding the LDAP commands that initiate a TLS connection:

The current examples in the guide:

```
ldapadd -x -ZZ -D cn=admin,dc=example,dc=com -W -f replicator.ldif
ldappasswd -x -ZZ -D cn=admin,dc=example,dc=com -W -S cn=replicator,dc=example,dc=com
```

These commands attempt to start a TLS connection but do not specify the LDAP server's hostname. This can cause the command to fall back to localhost, leading to TLS verification failures if the certificate is not issued for localhost.

To fix this and avoid TLS certificate mismatch errors, I suggest updating the examples to explicitly include the hostname:

```
ldapadd -x -ZZ  -H ldap://ldap01.example.com -D cn=admin,dc=example,dc=com -W -f replicator.ldif
ldappasswd -x -ZZ  -H ldap://ldap01.example.com -D cn=admin,dc=example,dc=com -W -S cn=replicator,dc=example,dc=com
```

This ensures that the correct hostname is used, preventing TLS certificate verification failures. This improvement comes from the following discussion: https://discourse.ubuntu.com/t/-/11578 (TLS guide).

Thanks for considering this!

Best regards,
Daniele Rugginenti

---
*Reported from: https://documentation.ubuntu.com/server/how-to/openldap/replication/*

PR #139, opened 2025-02-10
openldap replication: cleanup and tip about -H parameter

panlinux:
Use -H instead of -h as the former the preferred way to specify an LDAP server nowadays, as it's a full URI and incorporates TLS and other transport details:
- ldap:// for standard cleartext 389 port access
- ldaps:// for LDAP over TLS (port 636)
- ldapi:// for unix socket

Also add a tip saying that the default can be set in /etc/ldap/ldap.conf

Fixes: #65


s-makin (approved):
LGTM :)

PR #140, opened 2025-02-11
glossary: create 8 A-definitions

Sophie-Pages:
# Description

Create 8 definitions of terms starting with "A".

Partial canonical/open-documentation-academy#166

## Changes
- Create definitions
- Add missing terms in the glossary
- Link terms to the glossary the first time they appear on each page

## Questions or remarks
- Should we convert the glossary into MyST to uniform the pages? I am wondering why we choose to have the glossary in RST over MyST.
- I started defining a glossary convention: each definiton should explain what the term **is** first (except when it is an acronym, then it should be explained first). And each definition should end with "Related topic(s):" when there are related topic(s). 
- I don't think the term `accesslog` qualifies as a valid term but I wanted to double check: [documentation ](https://documentation.ubuntu.com/server/how-to/openldap/replicat)where it appear. If it is not a valid term, I will remove it from the glossary.

### DIFF
--- a/explanation/crypto/openssh-crypto-configuration.md
+++ b/explanation/crypto/openssh-crypto-configuration.md
@@ -99,7 +99,7 @@ For the other stages in the `ssh` connection, like key exchange, or public key a
 
 ### Remove AES 128 from server
 
-Let's configure an OpenSSH server to only offer the {term}`AES` 256 bit variant of symmetric ciphers for an `ssh` connection.
+Let's configure an OpenSSH server to only offer the {term}`AES` 256-bit variant of symmetric ciphers for an `ssh` connection.
 
 First, let's see what the default is:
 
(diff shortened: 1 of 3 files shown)


s-makin (commented):
Thanks for your work on this @Sophie-Pages! I really love the choices you made in your formatting of the glossary entries, especially the additional linking which will really help readers to navigate to what they need. 

With regards to the question you raised in the initial PR, I think you're right that it makes more sense for the page to be in Markdown/MyST format, but we will do a "final" pass on the document anyway to standardise all the differences in formatting once it's closer to completion. At that point, we can change it to markdown as well - there are some other pages that also need to be changed from rST to markdown to reduce the confusion from supporting both, so that can all be done at the same time.

I've answered your other questions inline, and left a couple more nits/suggestions, but otherwise this looks great!

s-makin (approved):
This is awesome, and thanks for also taking the extra step of adding missing terms to the glossary! :)

Sophie-Pages:
Thank you to you too @s-makin for your quick and useful review!

PR #190, opened 2025-03-09
about-apt-upgrade-and-phased-updates: remove note about closed bug

kenyon:
This bug has now been closed: https://bugs.launchpad.net/ubuntu/+source/apt/+bug/1988819


s-makin (approved):
Since we didn't actually link to the bug, and the bug is now fixed anyway, this seems like a sensible removal to me. Thanks for taking the time to submit this PR :)

PR #200, opened 2025-03-16
Clarify network interface configuration

Zetway-Technologies:
Updated instructions to specify the need for the correct network interface name (e.g., eth0, ens33).

### Description

Clarified the process for configuring the network interface in the `isc-dhcp-server` setup. 
 Updated the documentation to emphasize the importance of specifying the  ***correct network interface name*** in `/etc/default/isc-dhcp-server`. 

---


lvoytek (approved):
Thanks for adding this! This is a useful clarification

PR #201, opened 2025-03-17
Virtualisation architecture overhaul part 1

network-charles:
### Description

This PR addresses Part 1 of the Virtualisation docs overhaul. It contains the following justification for the changes:.

> When I mention a Line, I am referring to the old content.

#### Virtualisation with QEMU section

1. The contents in the 3 QEMU sections at the `how-to/virtualisation/` path are short and focus on a similar topic, QEMU; installation, adding vCPU, and installing the arm version. So, combining them into one page provides a complete reference to the core QEMU application, while still ensuring that this single page isn't too long.
2. The previous page `how-to/virtualisation/qemu` had the following issues and changes:
     - At Line 1, I updated the label from `(virtualisation-with-qemu)=` to `(qemu)=`. I also updated all references to the old label on these pages. This find was made possible using this Linux command: `find ./ -name "virtualisation-with-qemu"`
        - `explanation/virtualisation/about-openstack.md`
        - `explanation/virtualisation/vm-tools-in-the-ubuntu-space.md`
        - `how-to/virtualisation.rst`
     - Since the new label is `qemu`, I removed `virtualisation-with-qemu` in `redirects.txt` and updated it.
3. At Line 9, the original sentence lacked a clear contrast between QEMU’s built-in tools and libvirt. Adding "On the other hand" makes this distinction explicit, improving clarity.
4. At Line 11-16, the page also lacked a critical step, installing QEMU. Rather, it assumes that the reader already has QEMU installed. There was also no command to check if the system supports `kvm`. These steps align with the Diataxis how-to-guide framework.
5. I added a new section titled "Boot a VM" because it lets the user know that this is the next step after installing QEMU/KVM.
6. The netboot command does not work on headless KVM systems, I tried it on mine. So, I added a note letting new users know that if they are using a headless system, they can access the GUI by specifying VNC as their display method.
7. I added a new section called "Create a virtual disk" because it lets the user know that this is the next step after booting a VM. It also lets the user know that this step is optional but important. This helps improve Line 25, where the sentence said we can create a disk, but it didn't say why we are attaching one.
8. I removed the "Next steps" section at Line 38 and added a sentence about creating QEMU VMs with up to 1024 vCPUs, so the paragraphs flow smoothly into the next section.

#### Create QEMU VMs with up to 1024 vCPUs section

1. I deleted `how-to/virtualisation/qemu-vms-up-to-1024-vcpus.md` and added the page to `how-to/virtualisation/qemu` as I said initially.
   - Since the label `create-qemu-vms-with-up-to-1024-vcpus` is no longer available because the page has been deleted, I removed the label from the redirect page, `redirects.txt`.
   - I also removed the label `create-qemu-vms-with-up-to-1024-vcpus` from this page `how-to/virtualisation.rst`. It was the only page referencing that label. This find was made possible using this Linux command: `find ./ -name "create-qemu-vms-with-up-to-1024-vcpus"`
2. I changed the previous section title from "Supported QEMU versions" to use "Requirement" because the new word is more common in a how-to guide. I also changed the content of the requirement to a table so it's comprehensible.
3. I created a new section titled "Configuration by Ubuntu release" because this is more common in a how-to guide. The previous content went straight to the configuration for each release without letting a user skimming the page quickly know that this is the section around configuring based on Ubuntu release.

#### Boot ARM64 virtual machines on QEMU section

I updated the "note" so it aligns with the style guide. 

#### Resources section

This section allows the user navigate to QEMU-related pages.

#### QEMU microvm page

I improved a few grammars on the qemu-microvm page.

### Related Issue

Relevant to canonical/open-documentation-academy#185

### Checklist


how-to/virtualisation/qemu.md
```diff
@@ -1,38 +1,50 @@
...
 ```bash
 sudo apt-get install qemu-system
 ```
 
-While there *are* more user-friendly and comfortable ways, the quickest way to get started with QEMU is by booting a VM directly from a netboot ISO. You can achieve this by running the following command:
+### Boot a VM
 
-> **Warning**: 
-> This example is just for illustration purposes - it is not generally recommended without verifying the checksums; {ref}`Multipass <create-vms-with-multipass>` and {ref}`UVTool <cloud-image-vms-with-uvtool>` are much better ways to get actual guests easily.
+While there _are_ more user-friendly and comfortable ways like downloading the ISO provides for faster access at runtime, the quickest way to get started with QEMU is by booting a VM directly from a netboot ISO. You can achieve this by running the following command:
```
j5awry, line 33: cleanup and grammar. I suggest dropping the opening statement, as the below caution states other "user friendly" and secure options.

The quickest way to get started with QEMU is by booting a VM directly from a netboot ISO. You can achieve this by running the following command:

network-charles: Nice suggestion, thanks!

network-charles:
Hi @s-makin, I'm done with this part. I'd appreciate your feedback.

s-makin:
> Hi @s-makin, I'm done with this part. I'd appreciate your feedback.

Thanks @network-charles - I've put this in my queue for tomorrow (there are a couple of older PRs in my review queue that I need to address before this one). Are you still working on part 2?

network-charles:
Nice! Yes, I'm still working on it.

redirects.txt
```diff
@@ -45,11 +45,7 @@ install/lpar-autoinstall-on-s390x how-to/installation/non-interactive-ibm-z-lpar
 non-interactive-ibm-z-lpar-autoinstall-s390x how-to/installation/non-interactive-ibm-z-lpar-autoinstall-s390x/
 
 virtualization-qemu how-to/virtualisation/qemu/
-virtualisation-with-qemu how-to/virtualisation/qemu/
-
-create-qemu-vms-with-up-to-1024-vcpus how-to/virtualisation/qemu-vms-up-to-1024-vcpus/
-
-boot-arm64-virtual-machines-on-qemu how-to/virtualisation/arm64-vms-on-qemu/
```
s-makin, line 52: These redirects are sending traffic that comes from the old Discourse-based pages to the correct place in this documentation.

For that reason, we can't just remove the old redirects - we need to point them to the correct place, otherwise users are going to get 404s everywhere. You'll need to reinstate the removed/changed redirects and *add* the new ones. Redirect files only get bigger, never shorter :D

So what you'll need is something like:
```
virtualization-qemu how-to/virtualisation/qemu/
virtualisation-with-qemu how-to/virtualisation/qemu/

create-qemu-vms-with-up-to-1024-vcpus how-to/virtualisation/qemu-vms-up-to-1024-vcpus/
how-to/virtualisation/qemu-vms-up-to-1024-vcpus/ how-to/virtualisation/qemu

boot-arm64-virtual-machines-on-qemu how-to/virtualisation/arm64-vms-on-qemu/
how-to/virtualisation/arm64-vms-on-qemu/ how-to/virtualisation/qemu
```

I don't know if it's possible to redirect to sections within a page using the redirect file - I suspect it's not but it might be worth having a play around just to be sure (i.e. instead of redirecting the two pages to just `qemu`, try `qemu#arm64-vms-on-qemu`). If you can test that locally, that'd be great - it should work on your local preview.

explanation/virtualisation/qemu-microvm.md
```diff
@@ -42,19 +42,19 @@ There are a few common ways adapt a workload to match this:
   guest system.
 - Sometimes a simpler user space like [busybox](https://www.busybox.net/) or a container-like environment
   is used.
-- In a similar fashion, a customised kernel build with a reduced feature set
+- In a similar fashion, a customised kernel builds with a reduced feature set
```
s-makin, line 45: I think this is proposing "a customised kernel build with a reduced feature set" - rather than just a "customised kernel", which could have either a reduced or expanded feature set. It's not well phrased in the original, which makes it confusing. I'd suggest something like:
"In a similar fashion, you could use a customised kernel build with a reduced..."

Wdyt?

explanation/virtualisation/qemu-microvm.md
```diff
@@ -8,8 +8,8 @@ The underlying concept of a microvm is based on giving up some capabilities of
 standard QEMU in order to reduce complexity and gain speed. Maybe - for your
 use-case - you do not need e.g. the hypervisor to be able to pretend to have a
 network card from the 90s, nor to emulate a CPU of a foreign architecture,
-nor live migrate with external I/O going on. In such cases a lot of what QEMU
-provides is not needed and a less complex approach like microvm might be
+nor live migrate with external I/O going on. In such cases, a lot of what QEMU
+provides are not needed and a less complex approach like microvm might be
```
s-makin, line 12:
```suggestion
provides is not needed and a less complex approach like a microvm might be
```

s-makin: The original was correct :)

how-to/virtualisation/qemu.md
```diff
@@ -35,6 +60,147 @@ qemu-system-x86_64 options image[s]
 
 So take a look at the [QEMU manpage](http://manpages.ubuntu.com/manpages/bionic/man1/qemu-system.1.html), [`qemu-img`](http://manpages.ubuntu.com/manpages/bionic/man1/qemu-img.1.html) and the [QEMU documentation](https://www.qemu.org/documentation/) and see which options best suit your needs.
 
-## Next steps
+While a standard QEMU configuration works for most use cases, some scenarios demand high-vCPU VMs. In the next section, we’ll cover how to create QEMU virtual machines with up to 1024 vCPUs.
+
```
s-makin, line 64: It might be a good idea to keep the anchors from the original pages, above the original titles. In many cases this should ensure that any links to those original anchors don't get broken (any that are will be picked up by the redirects).

how-to/virtualisation/qemu.md
```diff
@@ -1,25 +1,50 @@
...
 [QEMU](http://wiki.qemu.org/Main_Page) is a machine emulator that can run operating systems and programs for one machine on a different machine. However, it is more often used as a virtualiser in collaboration with [KVM](https://www.linux-kvm.org/page/Main_Page) kernel components. In that case it uses the hardware virtualisation technology to virtualise guests.
 
-Although QEMU has a [command line interface](https://qemu-project.gitlab.io/qemu/system/invocation.html ) and a [monitor](https://qemu-project.gitlab.io/qemu/system/monitor.html) to interact with running guests, they are typically only used for development purposes. [libvirt]( libvirt.md) provides an abstraction from specific versions and hypervisors and encapsulates some workarounds and best practices.
+Although QEMU has a [command line interface](https://qemu-project.gitlab.io/qemu/system/invocation.html) and a [monitor](https://qemu-project.gitlab.io/qemu/system/monitor.html) to interact with running guests, they are typically only used for development purposes. On the other hand, [libvirt](libvirt.md) provides an abstraction from specific versions and hypervisors and encapsulates some workarounds and best practices.
```
s-makin, line 13:
```suggestion
Although QEMU has a [command line interface](https://qemu-project.gitlab.io/qemu/system/invocation.html) and a [monitor](https://qemu-project.gitlab.io/qemu/system/monitor.html) to interact with running guests, they are typically only used for development purposes. On the other hand, {ref}`libvirt <libvirt>` provides an abstraction from specific versions and hypervisors, and encapsulates some workarounds and best practices.
```

s-makin: Looks like I missed that link when I was converting from files to internal anchors.

how-to/virtualisation/qemu.md
```diff
@@ -1,25 +1,50 @@
...
+
+```bash
+kvm-ok
+```
 
-> **Warning**: 
-> This example is just for illustration purposes - it is not generally recommended without verifying the checksums; {ref}`Multipass <create-vms-with-multipass>` and {ref}`UVTool <cloud-image-vms-with-uvtool>` are much better ways to get actual guests easily.
+You should get an output saying `KVM acceleration can be used`.
+
+Then next step is to install QEMU.
```
s-makin, line 25:
```suggestion
The next step is to install QEMU.
```

how-to/virtualisation/qemu.md
```diff
@@ -35,6 +60,147 @@ qemu-system-x86_64 options image[s]
 
 So take a look at the [QEMU manpage](http://manpages.ubuntu.com/manpages/bionic/man1/qemu-system.1.html), [`qemu-img`](http://manpages.ubuntu.com/manpages/bionic/man1/qemu-img.1.html) and the [QEMU documentation](https://www.qemu.org/documentation/) and see which options best suit your needs.
 
-## Next steps
+While a standard QEMU configuration works for most use cases, some scenarios demand high-vCPU VMs. In the next section, we’ll cover how to create QEMU virtual machines with up to 1024 vCPUs.
```
s-makin, line 63: Love your ordering of the 2 additional pages here - this is a logical progression and the context you added makes this flow really nicely.

how-to/virtualisation/qemu.md
```diff
@@ -35,6 +60,147 @@ qemu-system-x86_64 options image[s]
...
+
+## Boot ARM64 virtual machines on QEMU
+
+Ubuntu ARM64 images can run inside QEMU. You can either do this fully emulated (e.g. on an x86 host) or accelerated with KVM if you have an ARM64 host. This page describes how to do both.
+
+```{note}
+This requires Ubuntu 20.04 or greater
+```
+
+### Install QEMU
```
s-makin, line 132: Can we rename this header to be a bit more specific? It looks weird to have "install QEMU" be a header this far down the page

how-to/virtualisation/qemu.md
```diff
@@ -35,6 +60,147 @@ qemu-system-x86_64 options image[s]
...
+
+### Troubleshooting
+
+#### No output and no response
+
+If you get no output from the QEMU command above, aligning your host and guest release versions may help. For example, if you generated `efi.img` on Focal but want to emulate Jammy (with the Jammy cloud image), the firmware may not be fully compatible. Generating `efi.img` on Jammy when emulating Jammy with the Jammy cloud image may help.
+
+## Resources
+
+QEMU can be extended in many different ways. If you'd like to take QEMU further, you might want to explore these.
```
s-makin, line 203:
```suggestion
QEMU can be extended in many different ways. If you'd like to take QEMU further, you might want to explore these additional resources:
```

how-to/virtualisation/qemu.md
```diff
@@ -35,6 +60,147 @@ qemu-system-x86_64 options image[s]
...
 
-QEMU can be extended in many different ways. If you'd like to take QEMU further, you might want to check out this follow-up guide on {ref}`virtualizing graphics using QEMU/KVM <gpu-virtualization-with-qemu-kvm>`, or this guide on how you can {ref}`use QEMU to create MicroVMs <qemu-microvm>`.
+- {ref}`Virtualizing graphics using QEMU/KVM <gpu-virtualization-with-qemu-kvm>`
+- {ref}`Using QEMU to create MicroVMs <qemu-microvm>`.
```
s-makin, line 206: Ah, I overlooked this originally as well. I believe QEMU microvms are not capitalised, so it should be microvm and not microVM

redirects.txt
network-charles: Ah, nice, I see it now. I had to check the URL on one of the discourse pages.

> I don't know if it's possible to redirect to sections within a page using the redirect file

I tried it, but it's not possible. The redirect points to an `index.html` file in the build, so including the section via a #, just makes it assume that the section is a path containing an `index.html` file.
The error message:
```bash
WARNING: (broken) boot-arm64-virtual-machines-on-qemu/index.html redirects to how-to/virtualisation/qemu#arm64-vms-on-qemu/index.html but /home/ubuntu/_build/how-to/virtualisation/qemu#arm64-vms-on-qemu/index.html does not exist!
WARNING: (broken) how-to/virtualisation/arm64-vms-on-qemu/index.html redirects to how-to/virtualisation/qemu#arm64-vms-on-qemu/index.html but /home/ubuntu/_build/how-to/virtualisation/qemu#arm64-vms-on-qemu/index.html does not exist!
```

how-to/virtualisation/qemu.md
network-charles: Ah, true!

explanation/virtualisation/qemu-microvm.md
network-charles: I agree, that part was tricky. I like your suggestion

how-to/virtualisation/qemu.md
network-charles: Alright

s-makin (approved):
Thanks for making the requested changes, all looks good to me! And thanks for investigating the redirect-to-a-subheading issue as well. I did have a feeling that wouldn't work, but it's good to have it confirmed.

I've taken the liberty of fixing up the last of the redirects - I'll keep an eye on the traffic as well just to make sure it's all working as expected after merging.

network-charles:
Nice! You are welcome

PR #251, opened 2025-04-24
Further landing page improvements

s-makin:
### Description

Part 8 of TBD? I've lost track a bit. 

This page standardises between similar topics in the how-to and explanation sections
and adds contextual text to all entries on the how-to landing page

### Checklist


bryceharrington (approved):
LGTM.

It looks like you've got a few distinct refactorings going on here; for future reference it can make review a bit easier if those are split out in separate commits so each change is clearer as to what it's doing - and thus quicker to catch errors.  But I gather this is an ongoing work so if there are any errors missed you'll catch them in the next PR.

s-makin:
> LGTM.
> 
> It looks like you've got a few distinct refactorings going on here; for future reference it can make review a bit easier if those are split out in separate commits so each change is clearer as to what it's doing - and thus quicker to catch errors. But I gather this is an ongoing work so if there are any errors missed you'll catch them in the next PR.

Thanks @bryceharrington - usually I do try to separate my commits, but I haven't bothered with these PRs because as you say, I'm making lots of changes in quick succession. I've got (probably) two more to go before I'm done. The next one will address the last pages that haven't been revised yet, and then the final one to catch any bits that still need tidying up. In terms of restructuring, I've gone as far as I'm going to go for now - I'm not planning to combine or move any further sections in this round (unless you have suggestions, which I'm happy to take!)

PR #272, opened 2025-05-23
Update user-management.md password policy to accurately reflect example

kajgm:
### Description

The docs reference an example where `/etc/pam.d/common-password` is modified to have a minimum length of 8 characters. This example references a `minlen=8` field when the explanation immediately above specifies the field as `min=8`.


lvoytek:
Thank you for the merge request! `minlen` is indeed correct and should be reflected in the description of the example. Merging in

PR #273, opened 2025-05-26
[Update] Remove mentions of no-longer maintained or recommended observability tools and stacks.

simskij:
Update the observability section to no longer talk about LMA, or mention outdated tooling. Instead, provide links to the COS documentation, and to the OpenTelemetry collector snap.

### DIFF
--- a/how-to/observability.rst
+++ b/how-to/observability.rst
@@ -5,5 +5,8 @@ Observability
 
 In Ubuntu, it is recommended to use the
 `Canonical Observability Stack <https://documentation.ubuntu.com/observability/>`_
-to monitor your infrastructure. For more information about how to do this, see
-`the documentation on how to integrate COS Lite with uncharmed applications <https://documentation.ubuntu.com/observability/how-to/integrating-cos-lite-with-uncharmed-applications/>`_.
+to monitor your infrastructure. For more information, see the following links:
+
+- `Getting started with COS Lite <https://documentation.ubuntu.com/observability/tutorial/installation/getting-started-with-cos-lite/>`_
+- `How to integrate COS Lite with uncharmed applications <https://documentation.ubuntu.com/observability/how-to/integrating-cos-lite-with-uncharmed-applications/>`_
+- `The OpenTelemetry Collector snap <https://snapcraft.io/opentelemetry-collector>`_


how-to/observability.rst
```diff
@@ -5,23 +5,8 @@ Observability
...
-to monitor your infrastructure.
+to monitor your infrastructure. For more information, see the following links:
 
-However, you can also use the classic Logging, Monitoring and Alerting (LMA)
-stack.
-
-.. toctree::
-   :titlesonly:
-
-   Set up your LMA stack <observability/set-up-your-lma-stack>
```
s-makin, line 16: I'm happy for us to take away the LMA stack, since this has been fully superseded by COS. However, I would ask you to also delete the lma stack page in question, otherwise the documentation will fail to build because of the page existing without being in a toctree.

how-to/observability.rst
```diff
@@ -5,23 +5,8 @@ Observability
...
-.. toctree::
-   :titlesonly:
-
-   Install Logwatch <observability/install-logwatch>
-   Install Munin <observability/install-munin>
-   Install Nagios Core 3 <observability/install-nagios>
-   Use Nagios with Munin <observability/use-nagios-with-munin>
+- `Getting started with COS Lite <https://documentation.ubuntu.com/observability/tutorial/installation/getting-started-with-cos-lite/>`_
+- `How to integrate COS Lite with uncharmed applications <https://documentation.ubuntu.com/observability/how-to/integrating-cos-lite-with-uncharmed-applications/>`_
+- `The OpenTelemetry Collector snap <https://snapcraft.io/opentelemetry-collector>`_
```
s-makin, line 12: These links are a really helpful addition, but if we want to tempt people to use COS we should provide a bit more context around these links.

I think if we can use this space to disambiguate the difference between the COS and COS lite use cases, and say what OTC does and what use case it is for, it will help our readers to understand which of the options they should use for their situation, which will in turn help increase the adoption.

s-makin (changes requested):
Thanks @simskij for submitting these changes. I do have some suggestions which will need to be addressed before I can merge.

This is a good place to ...not *advertise* exactly, but rather to explain the benefits of COS/COS lite and OTC and what makes them so useful for a sysadmin. Most of our readers already have established setups with tools they've used before, so if we want them to switch we need to tell them why they should. We've done this a few times in the Server docs, for example with [Netplan](https://documentation.ubuntu.com/server/explanation/networking/about-netplan/) and [cloud-init](https://documentation.ubuntu.com/server/explanation/intro-to/cloud-init/) which mostly exists to link to the official documentation. This might be out of scope for this PR, but I would encourage you to think about it as a possible follow up - we can discuss in the next observability sync if you want.

My other suggestions are inline :)

blg-666 (approved):

blg-666 (approved):

PR #282, opened 2025-06-02
Update qemu-microvm.md

raviksharma:
Fixing rock reference


cpaelzer:
Thank you, that is correct - the old link is now a 404 - you picked the right new link
Approve and merge ...

PR #295, opened 2025-06-15
Small improvements to PostgreSQL installation guide

simonharris:
### Description

* Correction in intro regarding authentication method
* Clarification to note regarding listen_addresses
* Use more common IP range in example

---

### Checklist

- [X] I have read and followed the [Ubuntu Server contributing guide](https://documentation.ubuntu.com/server/contributing/).
- [X] I have signed the [Contributor License Agreement (CLA)](https://ubuntu.com/legal/contributors).

### DIFF
--- a/how-to/databases/install-postgresql.md
+++ b/how-to/databases/install-postgresql.md
@@ -17,7 +17,7 @@ The database service is automatically configured with viable defaults, but can b
 
 PostgreSQL supports multiple client authentication methods. In Ubuntu, `peer` is the default authentication method used for `local` connections, while `scram-sha-256` is the default for `host` connections (this used to be `md5` until Ubuntu 21.10). Please refer to the [PostgreSQL Administrator's Guide](http://www.postgresql.org/docs/current/static/admin.html) if you would like to configure alternatives like Kerberos.
 
-The following discussion assumes that you wish to enable TCP/IP connections and use the MD5 method for client authentication. PostgreSQL configuration files are stored in the `/etc/postgresql/<version>/main` directory. For example, if you install PostgreSQL 14, the configuration files are stored in the `/etc/postgresql/14/main` directory.
+The following discussion assumes that you wish to enable TCP/IP connections and use the `scram-sha-256` method for client authentication. PostgreSQL configuration files are stored in the `/etc/postgresql/<version>/main` directory. For example, if you install PostgreSQL 14, the configuration files are stored in the `/etc/postgresql/14/main` directory.
 
 ```{tip}
 To configure *IDENT* authentication, add entries to the `/etc/postgresql/*/main/pg_ident.conf` file. There are detailed comments in the file to guide you.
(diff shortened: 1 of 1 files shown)


how-to/databases/install-postgresql.md
```diff
@@ -30,7 +30,7 @@ listen_addresses = '*'
 ```
 
 ```{note}
-'*' will allow all available IP interfaces (IPv4 and IPv6), to only listen for IPv4 set `0.0.0.0` while '`::`' allows listening for all IPv6 addresses.
+'`*`' will allow connections from all available IP interfaces (IPv4 and IPv6). To only listen for IPv4 set `listen_addresses` to '`0.0.0.0`', while '`::`' allows listening for all IPv6 addresses.
```
panlinux, line 33: This is still confusing (as was the original text). Let's unpack it.

First, `listen_addresses`  is not about "allowing" connections "from" all available IP interfaces. This is about where the postgresql daemon will listen on.

So I would say that `*` will have postgresql listening on all available network interfaces on that system (IPv4 or IPv6, doesn't matter). The distinction about IPv4 and IPv6 sounds correct, and your text is better.

how-to/databases/install-postgresql.md
```diff
@@ -41,16 +41,16 @@ Now that we can connect to our PostgreSQL server, the next step is to set a pass
...
 ```bash
 ALTER USER postgres with encrypted password 'your_password';
 ```
 
-After configuring the password, edit the file `/etc/postgresql/*/main/pg_hba.conf` to use `scram-sha-256` authentication with the `postgres` user, allowed for the `template1` database, from any system in the local network (which in the example is `192.168.122.1/24`) :
+After configuring the password, edit the file `/etc/postgresql/*/main/pg_hba.conf` to use `scram-sha-256` authentication with the `postgres` user, allowed for the `template1` database, from any system in the local network (which in the example is `192.168.1.1/24`) :
 
 ```text
-hostssl template1       postgres        192.168.122.1/24        scram-sha-256
+hostssl template1       postgres        192.168.1.1/24        scram-sha-256
```
panlinux, line 53: I can see why `192.168.122.1/24` was used, that's the default network that you get with a libvirt-based VM. It's not wrong. I guess your point is that `192.168.1.1/24` is more common, but do you have any other issue with this example>

panlinux (changes requested):
The `listen_addresses` explanation needs a bit of work, otherwise this is ok.

how-to/databases/install-postgresql.md
simonharris: I stand by the change. I accept that the original isn't strictly "wrong" per se, at least in the context of VMs, but this documentation is not about configuring VMs. 

Using a range that only applies in a specific context brings no benefit, and I believe that using the more likely 191.168.1.1/24 carries less of a surprise factor for the reader.

panlinux (approved):
+1, thanks for the PR and the updates

Issue #216, opened 2025-03-31
[Feedback]: OpenVPN guide on 24.04


valerio-bozzolan:
### Page URL

https://documentation.ubuntu.com/server/how-to/security/install-openvpn/index.html

### Select an option

- [ ] I found what I was looking for
- [ ] I couldn't find what I was looking for
- [x] I found the information, but it was incorrect
- [ ] I found the information, but it was incomplete
- [ ] I found the information, but it was confusing
- [ ] I encountered a technical issue (e.g., broken link, image not loading)

## Feedback 1

The mentioned guide cite this step:

    sudo cp /usr/share/doc/openvpn/examples/sample-config-files/server.conf.gz /etc/openvpn/myserver.conf.gz

But in Ubuntu 24.04 that file is not compressed.

So I think it's just this:

    sudo cp /usr/share/doc/openvpn/examples/sample-config-files/server.conf /etc/openvpn/myserver.conf

## Feedback 2

The guide mentions this step:

    sudo openvpn --genkey --secret ta.key

But it generates this:

    2025-03-31 09:29:16 DEPRECATED OPTION: The option --secret is deprecated.
    2025-03-31 09:29:16 WARNING: Using --genkey --secret filename is DEPRECATED.  Use --genkey secret filename instead.

Accordingly to the man page, this seems the correct version instead:

    sudo openvpn --genkey secret ta.key

valerio-bozzolan:
If a merge request is welcome I can try.

slyon:
> If a merge request is welcome I can try.

Yes a PR would be very welcome!
And we probably also want to preserve the pre-24.04 commands. We're still discussing how we'd like to implement this. So for now I'd say, keep the pre-24.04 information around, but don't get fancy about it, as the way that is presented will probably change in the short term.

valerio-bozzolan:
Appreciated ping. If I don't do this after this Monday, consider me dead lol

valerio-bozzolan:
Aaaand here we are @panlinux 👍 #296

(If necessary feel free to assign this issue to me, so others don't spend time on this. I cannot self-assign lol)

valerio-bozzolan:
Plus, I just noticed that official Ubuntu Docker images have not the path `/usr/share/doc` so half documentation cannot be followed.

Maybe in the troubleshooting section we can also mention that if `/usr/share/doc` is missing, they need to update this file (or a similar one) and re-install `openvpn`:

```
[...]

PR #296, opened 2025-06-15
OpenVPN: fix guide for Ubuntu 24.04+

valerio-bozzolan:
### Description

Fix all the small issues described in #216 for OpenVPN in Ubuntu 24.04.

- Fix un-existing server.conf.gz
- Fix deprecated warning «`DEPRECATED OPTION: The option --secret is deprecated.`»
- Fix missing TLS Auth on the server

---

### Related Issue

- Fixes: #216

---

### Merge Strategy

I suggest to do **not** squash the commits, since each commit has meaningful details for the benefit of future 'git blame'. Therefore I suggest to merge as **fast-forward** (or merge commit). You're welcome! :+1: lol

---

### Checklist

- [X] I have read and followed the [Ubuntu Server contributing guide](https://documentation.ubuntu.com/server/contributing/).
- [X] I have signed the [Contributor License Agreement (CLA)](https://ubuntu.com/legal/contributors).

### DIFF
--- a/how-to/security/install-openvpn.md
+++ b/how-to/security/install-openvpn.md
@@ -106,7 +106,16 @@ total 68
 -rw-r--r-- 1 root root 4141 2011-07-04 15:09 server.conf.gz
 ```
 
-Start by copying and unpacking `server.conf.gz` to `/etc/openvpn/server.conf`:
+Start by copying and unpacking `server.conf.gz` to `/etc/openvpn/server.conf`.
+
+In Ubuntu 24.04 or newer:
+
+```bash
+sudo cp /usr/share/doc/openvpn/examples/sample-config-files/server.conf /etc/openvpn/myserver.conf
+sudo gzip -d /etc/openvpn/myserver.conf.gz
+```
+
+In Ubuntu 23.04 or older:
 
 ```bash
 sudo cp /usr/share/doc/openvpn/examples/sample-config-files/server.conf.gz /etc/openvpn/myserver.conf.gz
--- a/how-to/security/install-openvpn.md
+++ b/how-to/security/install-openvpn.md
@@ -350,6 +350,7 @@ If the above didn't work for you, check the following:
 - Client and server must use same protocol and port, e.g. UDP port 1194, see `port` and `proto` config options.
 - Client and server must use the same compression configuration, see `comp-lzo` config option.
 - Client and server must use same config regarding bridged vs. routed mode, see `server vs server-bridge` config option
+- Client must use the config `tls-auth` with index `0` (example client config: `tls-auth FILE 0`), but server must use `tls-auth` with index `1` (example server config: `tls-auth FILE 1`).
 
 ## Advanced configuration
 
--- a/how-to/security/install-openvpn.md
+++ b/how-to/security/install-openvpn.md
@@ -129,6 +129,7 @@ ca ca.crt
 cert myservername.crt
 key myservername.key
 dh dh.pem
+tls-auth ./easy-rsa/ta.key 0
 ```
 
 Complete this set with a TLS Authentication (TA) key in `etc/openvpn` for `tls-auth` like this:
--- a/how-to/security/install-openvpn.md
+++ b/how-to/security/install-openvpn.md
@@ -135,7 +135,7 @@ tls-auth ./easy-rsa/ta.key 0
 Complete this set with a TLS Authentication (TA) key in `etc/openvpn` for `tls-auth` like this:
 
 ```bash
-sudo openvpn --genkey --secret ta.key
+sudo openvpn --genkey secret ta.key
 ```
 
 Edit `/etc/sysctl.conf` and uncomment the following line to enable IP forwarding:


how-to/security/install-openvpn.md
```diff
@@ -106,7 +106,16 @@ total 68
...
+Start by copying and unpacking `server.conf.gz` to `/etc/openvpn/server.conf`.
+
+In Ubuntu 24.04 or newer:
+
+```bash
+sudo cp /usr/share/doc/openvpn/examples/sample-config-files/server.conf /etc/openvpn/myserver.conf
+sudo gzip -d /etc/openvpn/myserver.conf.gz
+```
+
+In Ubuntu 23.04 or older:
```
panlinux, line 118: We try to focus only on LTS releases for the documentation, so the above should use 22.04 instead of 23.04.

how-to/security/install-openvpn.md
```diff
@@ -106,7 +106,16 @@ total 68
...
 ```
 
-Start by copying and unpacking `server.conf.gz` to `/etc/openvpn/server.conf`:
+Start by copying and unpacking `server.conf.gz` to `/etc/openvpn/server.conf`.
+
+In Ubuntu 24.04 or newer:
+
+```bash
+sudo cp /usr/share/doc/openvpn/examples/sample-config-files/server.conf /etc/openvpn/myserver.conf
+sudo gzip -d /etc/openvpn/myserver.conf.gz
```
panlinux, line 115: This is copying an uncompressed file, so there is no need for the gzip call.

That being said, I just deployed ubuntu 22.04 and 24.04, and I see that both ship server.conf uncompressed, is that what you see too?

In 22.04, which has openvpn `2.5.11-0ubuntu0.22.04.1`:
```
ubuntu@j:~$ grep ^VERSION= /etc/os-release 
VERSION="22.04.5 LTS (Jammy Jellyfish)"
ubuntu@j:~$ ls -la /usr/share/doc/openvpn/examples/sample-config-files/server.conf
-rw-r--r-- 1 root root 10852 Sep 17  2024 /usr/share/doc/openvpn/examples/sample-config-files/server.conf
ubuntu@j:~$ 
```

And in 24.04, which has openvpn `2.6.12-0ubuntu0.24.04.3`:
```
ubuntu@n:~$ grep ^VERSION= /etc/os-release
VERSION="24.04.2 LTS (Noble Numbat)"
ubuntu@n:~$ ls -la /usr/share/doc/openvpn/examples/sample-config-files/server.conf
-rw-r--r-- 1 root root 10775 Apr  1 16:06 /usr/share/doc/openvpn/examples/sample-config-files/server.conf
ubuntu@n:~$ 
```

how-to/security/install-openvpn.md
```diff
@@ -120,12 +129,13 @@ ca ca.crt
 cert myservername.crt
 key myservername.key
 dh dh.pem
+tls-auth ./easy-rsa/ta.key 0
 ```
 
 Complete this set with a TLS Authentication (TA) key in `etc/openvpn` for `tls-auth` like this:
```
panlinux, line 135: I think this directory is missing a `/`, i.e., `/etc/openvpn` (not your change, but since we are here...)

how-to/security/install-openvpn.md
```diff
@@ -120,12 +129,13 @@ ca ca.crt
 cert myservername.crt
 key myservername.key
 dh dh.pem
+tls-auth ./easy-rsa/ta.key 0
```
panlinux, line 132: Why the `./easy-rsa/` directory prefix? The `ta.key` file is being generated in `/etc/openvpn` by the instructions further down. I think this block should look just like the client config further down, with the `0` replaced with `1` of course.

how-to/security/install-openvpn.md
```diff
@@ -340,6 +350,7 @@ If the above didn't work for you, check the following:
 - Client and server must use same protocol and port, e.g. UDP port 1194, see `port` and `proto` config options.
 - Client and server must use the same compression configuration, see `comp-lzo` config option.
 - Client and server must use same config regarding bridged vs. routed mode, see `server vs server-bridge` config option
+- Client must use the config `tls-auth` with index `1` (example client config: `tls-auth FILE 1`), but server must use `tls-auth` with index `0` (example server config: `tls-auth FILE 0`).
```
panlinux, line 353: Great addition, and I think you can safely replace `FILE` with `ta.key` here.

panlinux (changes requested):
Thanks a lot for this! Comments in the diff, some questions, some requests for changes.

TheJJ (changes requested):
thanks for your contribution! i've added an improvement idea, apart from that i think this is ready to ship :ship:

how-to/security/install-openvpn.md
```diff
@@ -100,13 +100,28 @@ Since the client certificates and keys are only required on the client machine,
...
-Start by copying and unpacking `server.conf.gz` to `/etc/openvpn/server.conf`:
+If these files under `/usr/share/doc/*` are not available:
+
+> Official minimal environments like Ubuntu Docker are optimized to never install stuff like documentation files (you probably also have not `sudo` there, etc).
+> To re-install the necessary documentation, at this point you can run these commands, as the root user:
+>
+> ```bash
+> echo 'path-include=/usr/share/doc/openvpn/examples/*' > /etc/dpkg/dpkg.cfg.d/my-openvpn
+> apt install --reinstall openvpn
+> ```
```
panlinux, line 116: I don't think this is the right place for this information. Nowhere else in the server guide do we talk about how to tweak such minimal environments when describing a particular service. If we want to go down this route, I believe it's better to document such steps in a separate entry in the server guide, perhaps near the installation docs, or even a new one that talks about minimal images.

how-to/security/install-openvpn.md
```diff
@@ -100,13 +100,28 @@ Since the client certificates and keys are only required on the client machine,
...
+
+Start by copying the example server configuration to `/etc/openvpn/server.conf`:
+
+```bash
+sudo cp /usr/share/doc/openvpn/examples/sample-config-files/server.conf /etc/openvpn/myserver.conf
+```
+
+In Ubuntu 20.04 or older, do this instead:
```
panlinux, line 124: We don't need to worry about 20.04. If it's different in 22.04, then it's worth adding a note. But if both 22.04 and 24.04 have the same file and can be used with the same set of instructions, then even better (and I think that's the case, the file is uncompressed in 22.04 and 24.04, right?).

panlinux (changes requested):
Thanks for the update! I think the only remaining point is the server.conf/client.conf compression. If 22.04 and 24.04 are the same, then we just need one set of instructions, and don't worry about 20.04 and earlier.

panlinux:
Lots of updates in this PR, nice! Taking a look!

how-to/security/install-openvpn.md
panlinux: I think suggesting to change your ubuntu installation just to fetch these sample config files is overkill. How about a link to fetch these from the internet? We could link directly to the package git repository. For example, those example files can be seen here: https://git.launchpad.net/ubuntu/+source/openvpn/tree/sample/sample-config-files?h=applied/ubuntu/noble-devel

We can link directly:
* [client.conf](https://git.launchpad.net/ubuntu/+source/openvpn/tree/sample/sample-config-files/client.conf?h=applied/ubuntu/noble-devel)
* [server.conf](https://git.launchpad.net/ubuntu/+source/openvpn/tree/sample/sample-config-files/server.conf?h=applied/ubuntu/noble-devel)

These are stable links guaranteed to always point at these files from the latest release of openvpn for Ubuntu Noble 24.04.

how-to/security/install-openvpn.md
panlinux: Yes.

panlinux:
Hi @valerio-bozzolan, are you still interested in working on this PR?

how-to/security/install-openvpn.md
panlinux: I'm going to merge this as is, and make changes in a follow-up branch.

PR #299, opened 2025-06-17
Failed to listen, sssd-nss.socket

psflachglas:
Failed to listen, sssd-nss.socket, sssd-pam.socket fix source:
https://bugs-devel.debian.org/cgi-bin/bugreport.cgi?bug=977462


s-makin (approved):
LGTM :) thanks for improving the documentation

PR #309, opened 2025-06-27
openssh-server: recommend ed25519 key type

TheJJ:
use https://ed25519.cr.yp.to/

### DIFF
--- a/how-to/security/openssh-server.md
+++ b/how-to/security/openssh-server.md
@@ -8,7 +8,7 @@ OpenSSH is a freely available version of the Secure Shell (SSH) protocol family 
 
 The OpenSSH server component, `sshd`, listens continuously for client connections from any of the client tools. When a connection request occurs, `sshd` sets up the correct connection depending on the type of client tool connecting. For example, if the remote computer is connecting with the SSH client application, the OpenSSH server sets up a remote control session after authentication. If a remote user connects to an OpenSSH server with `scp`, the OpenSSH server daemon initiates a secure copy of files between the server and client after authentication.
 
-OpenSSH can use many authentication methods, including plain password, public key, and Kerberos tickets.
+OpenSSH can use many authentication methods, including plain password, public key cryptography, and Kerberos tickets.
 
 ## Install OpenSSH
 
@@ -31,7 +31,9 @@ To configure the default behavior of the OpenSSH server application, `sshd`, edi
 There are many directives in the `sshd` configuration file, which control things like communication settings and authentication modes. The following are examples of configuration directives that can be changed by editing the `/etc/ssh/sshd_config` file.
 
 ````{tip}
-Before editing the configuration file, you should make a copy of the original `/etc/ssh/sshd_config` file and protect it from writing so you will have the original settings as a reference and to reuse as necessary. You can do this with the following commands:
+You can use {ref}`etckeeper <install-etckeeper>` to track changes in your `/etc/` with `git`.
+
+Alternatively, before editing the configuration file, you should make a copy of the original `/etc/ssh/sshd_config` file and protect it from writing so you will have the original settings as a reference and to reuse as necessary. You can do this with the following commands:
 
 ```bash
 sudo cp /etc/ssh/sshd_config /etc/ssh/sshd_config.original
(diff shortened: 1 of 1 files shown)


how-to/security/openssh-server.md
```diff
@@ -65,35 +67,57 @@ Many other configuration directives for `sshd` are available to change the serve
 
 ## SSH keys
 
-SSH allows authentication between two hosts without the need of a password. SSH key authentication uses a **private key** and a **public key**.
+SSH allows authentication between two hosts without the need of a password, using cryptographic keys instead.
+{term}`SSH-key` authentication uses a **private key** and a **public key**.
 
-To generate the keys, run the following command:
+To generate a key pair, run the following command:
```
panlinux, line 73: Maybe just expand this sentence a bit, to say which type of algorithm will be used. Suggestion:
`To generate a key pair using the Ed25519 Elliptic Curve algorithm, run the following command:`

Then down below, perhaps instead of saying "If needed", just say "For an RSA key, do ...". If there are good resources out there explaining at a high level the differences, that could also be linked here.

how-to/security/openssh-server.md
```diff
@@ -65,35 +67,57 @@ Many other configuration directives for `sshd` are available to change the serve
...
 
 ```bash
-chmod 600 .ssh/authorized_keys
+sudo journalctl -eu sshd.service
+```
+
+
+## Connection multiplexing
+
+To reduce connection times, multiple SSH sessions to the same `target_machine` can work over one {term}`TCP` connection.
```
panlinux, line 108: suggestion: "To reduce initial connection times", or "To reduce connection setup times".

TheJJ:
Thanks for your review, I've adjusted the text to hopefully cover both key types (filenames) and reasoning better!

how-to/security/openssh-server.md
```diff
@@ -65,35 +67,61 @@ Many other configuration directives for `sshd` are available to change the serve
...
-To generate the keys, run the following command:
+To generate a key pair, run the following command:
 
 ```bash
-ssh-keygen -t rsa
+ssh-keygen -t ed25519
 ```
 
-This will generate the keys using the **RSA Algorithm**. At the time of this writing, the generated keys will have 3072 bits. You can modify the number of bits by using the `-b` option. For example, to generate keys with 4096 bits, you can use:
+During creation, you will be prompted for a key password, which you would enter to use that key (every time, or instead using an [ssh agent](#ssh-agent).
```
panlinux, line 82: s/key password/key passphrase/

Or just s/key password/passphrase/, to avoid repeating "key" twice.

how-to/security/openssh-server.md
```diff
@@ -65,35 +67,61 @@ Many other configuration directives for `sshd` are available to change the serve
...
 ```
 
-Finally, double check the permissions on the `authorized_keys` file -- only the authenticated user should have read and write permissions. If the permissions are not correct then change them by:
+You should now be able to SSH to the `target_machine` without being prompted for a password.
+
+To troubleshoot this, have a look at the `target_machine`'s journal:
 
 ```bash
-chmod 600 .ssh/authorized_keys
+sudo journalctl -eu sshd.service
```
panlinux, line 103: It's `ssh.service`. Even though the `ssh.service` unit has an `sshd.service` alias, it doesn't seem to work for journalctl.

Maybe a better troubleshooting command would perhaps be `sudo journalctl -u ssh.service -f`? Or in addition. To see live logs.

how-to/security/openssh-server.md
```diff
@@ -65,35 +67,61 @@ Many other configuration directives for `sshd` are available to change the serve
...
+```
+
+Since you can have multiple {term}`SSH-key`s, you can configure which to use for which target machine in `.ssh/config` (see [`man ssh_config`](https://manpages.ubuntu.com/manpages/latest/en/man5/ssh_config.5.html)).
```
panlinux, line 106: Missing `~`: `~/.ssh/config`

And for the manpage, can we use the new manpage reference scheme here? I forgot the syntax, but should be similar to the `{term}` one.

how-to/security/openssh-server.md
```diff
@@ -65,35 +67,61 @@ Many other configuration directives for `sshd` are available to change the serve
...
+
+
+
+## Connection multiplexing
+
+To reduce connection setup times, multiple SSH sessions to the same `target_machine` can work over one {term}`TCP` connection.
+
+Configuration to activate multiplexing in `~/.ssh/config`:
+```conf
```
panlinux, line 115: Here, and below again, `conf` is not a valid "lexer":
`openssh-server.md:115: WARNING: Pygments lexer name 'conf' is not known [misc.highlighting_failure]`. Maybe just use `text`.

how-to/security/openssh-server.md
```diff
@@ -65,35 +67,61 @@ Many other configuration directives for `sshd` are available to change the serve
...
-To generate the keys, run the following command:
+To generate a key pair, run the following command:
 
 ```bash
-ssh-keygen -t rsa
+ssh-keygen -t ed25519
 ```
 
-This will generate the keys using the **RSA Algorithm**. At the time of this writing, the generated keys will have 3072 bits. You can modify the number of bits by using the `-b` option. For example, to generate keys with 4096 bits, you can use:
+During creation, you will be prompted for a key passphrase, which you would enter to use that key (every time, or instead using an {manpage}`ssh_agent(1)`.
```
panlinux, line 82: it's `ssh-agent(1)` (hyphen, not underscore). The underscore one leads to a 404 essentially.

panlinux (approved):
awesome, +1

PR #376, opened 2025-10-15
Adjust copybutton config

s-makin:
### Description

The Sphinx copybutton can be configured to strip prompts so that when a user copies a command block it only copies the command. This PR implements that behaviour so we can now include `$` and `#` in commands: now when a user copies a code block via the button the prompt and any lines without a prompt will be excluded.

(thanks to rkratky for the regex)

### Checklist

---

I tested this locally using the explanation > performance > PGO page, which contains code blocks including prompts.

### DIFF
--- a/custom_conf.py
+++ b/custom_conf.py
@@ -255,6 +255,20 @@ custom_html_js_files = [
     'js/bundle.js',
 ]
 
+# Configure copybutton extension
+# This option excludes line numbers and prompts from being selected when
+# users copy commands using the copybutton
+# https://sphinx-copybutton.readthedocs.io/en/latest/use.html#automatic-exclusion-of-prompts-from-the-copies
+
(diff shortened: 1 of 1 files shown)


athos-ribeiro:
> How does this work if we are showing the contents of a script?

The plugin only operates on blocks where it finds occurrences of the defined prompt (which is some regex based heuristics). Assuming we will give up trying to parse `#` as a prompt, scripts should be fine as long as there are no lines starting with `$ ` or foo@bar$ `.

There may be corner cases, but I cannot think of many ATM. AFAICT, there are no widely adopted languages where `$` is used as a comment delimiter. Other cases could include line breaks to specific stirings such as

```
some_command --param-receiving-string "\
foo@bar$ \
"
```

which looks odd for a script

PR #502, opened 2026-01-19
Skip dot in index.pl to avoid making it an URL

athos-ribeiro:
This is currently being converted into an URL (unintentionally)


s-makin (approved):
lgtm:)

PR #505, opened 2026-01-19
Apply minor improvements to the web servers page

athos-ribeiro:
* Do not say X is the most used system for Y without backing data
* Expand LAMP


s-makin (approved):
lgtm!

Issue #476, opened 2026-01-19
[Suggestions]: Improvements to networking key concepts


s-makin:
### Page URL

https://documentation.ubuntu.com/server/explanation/networking/networking-key-concepts/

### Select an option

- [ ] I found what I was looking for
- [ ] I couldn't find what I was looking for
- [ ] I found the information, but it was incorrect
- [ ] I found the information, but it was incomplete
- [ ] I found the information, but it was confusing
- [x] I encountered a technical issue (e.g., broken link, image not loading)

### Issue Description

* IP address: differentiate between IPv4 & IPv6
* Replace classful networks with CIDR (Classless Inter-Domain Routing)
* Mention OSPF alongside RIP for routing
* Level3/Lumen vs Verizon vs Quad9 & Cloudflare
* `/etc/network/interfaces` vs `/etc/netplan/`

PR #520, opened 2026-01-20
explanation:networking:key-concepts: Update to modern tooling, protocols and notation

slyon:
This PR addresses issue #476 by:

* differentiating between IPv4 and IPv6 addresses
* Switching to modern CIDR notation, instead of classful networking
* Mentioning the OSPF routing protocol in addition to legacy RIP
* Clarify public DNS nameservers, such as QUad9 or Cloudflare
* Document /etc/netplan as the primary entry-point for network configuration (instead of /etc/network/interfaces - aka ifupdown, which is deprecated as of 18.04 LTS)

Fixes: https://github.com/canonical/ubuntu-server-documentation/issues/476


s-makin (approved):
Thanks for the additional improvements, this is awesome :)

Issue #544, opened 2026-01-21
Rework for 26.04: kdump-tools


cpaelzer:
Here is a rough collection of my review of https://documentation.ubuntu.com/server/how-to/software/kernel-crash-dump/#kernel-crash-dump to be up to date and ready for the coming LTS

- It is default since 24.10 as mentioned drop mentioning 16.04 further below as we do not need or want to mention the old back and forth of defaults (one is on install one is with package installed - so just remove the 16.04 words)

- make sense of cmdline `crashkernel=2G-4G:320M,4G-32G:512M,32G-64G:1024M,64G-128G:2048M,128G-:4096M` where it is already explained mention this to be the current default and the example being shortened to explain itbetter

- Add troubleshooting: if memory can't be allocated e.g. due to a shortage it will fail to load, if unsure check the status of "systemctl status kdump-tools" which might show e.g.
Jan 21 06:59:11 r-vm kdump-tools[5348]: Memory for crashkernel is not reserved
Jan 21 06:59:11 r-vm kdump-tools[5348]: Please reserve memory by passing"crashkernel=Y@X" parameter to kernel
Jan 21 06:59:11 r-vm kdump-tools[5348]: Then try to loading kdump kernel
For example in a system with only 1G memory it considers itself too muhc of a burden relative to the overall resources. One can tweak that, but by default it tries to be helpful yet not negatively impactful.
Same answer on `kdump-config load` which you'd use after config change.

- USE_KDUMP=1 is the default since the mentioned 24.10 - the text needs to change as it no more needs to be enabled

- update example output of `kdump-config show`

bad: 

root@r-vm:~# kdump-config show
DUMP_MODE:                kdump
USE_KDUMP:                1
KDUMP_COREDIR:                /var/crash
crashkernel addr: 
   /var/lib/kdump/vmlinuz: symbolic link to /boot/vmlinuz-6.18.0-8-generic
kdump initrd: 
   /var/lib/kdump/initrd.img: symbolic link to /var/lib/kdump/initrd.img-6.18.0-8-generic
current state:    Not ready to kdump

crashkernel suggested size: 329M

kexec command:
  no kexec command recorded

good

kdump-config show
DUMP_MODE:                kdump
USE_KDUMP:                1
KDUMP_COREDIR:                /var/crash
crashkernel addr: 0x60000000
   /var/lib/kdump/vmlinuz: symbolic link to /boot/vmlinuz-6.18.0-8-generic
kdump initrd: 
[...]

PR #565, opened 2026-01-21
26.04 readiness kdump tools

cpaelzer:
Trying to address all review feedback found in #544

### DIFF
--- a/docs/how-to/software/kernel-crash-dump.md
+++ b/docs/how-to/software/kernel-crash-dump.md
@@ -55,13 +55,13 @@ During the installation, you will be prompted with the following dialog:
 
 ## `kdump` is enabled by default where applicable
 
-On manual install the interactive question will define if it is enabled or not
-and therefore depend on the users choice, but in cases when it is automatically
+On manual installations, the interactive question will define if it is enabled or not
+and therefore depends on the users choice. In cases where it is automatically
 preinstalled there is a balance to strike.
 The tool wants to be helpful by default, because when the crash happens it is
 too late to enable crash dump functionality.
-But at the same time it needs to reserve memory for the crash kernel to operate
-and that might have too much impact to very small systems as well as consume
+At the same time, it needs to reserve memory for the crash kernel to operate,
+and that might have too much impact on very small systems as well as consume
 a lot of memory on systems with many devices.
 Therefore starting in Oracular Oriole (24.10), the kernel crash dump facility
 will be enabled by default during standard Ubuntu Desktop or Ubuntu Server
@@ -100,7 +100,7 @@ a reboot will be required in order to reconsider the `crashkernel= boot` paramet
 
 ## Verify if kdump is ready
 
-Upon reboot, the crashkernel memory will be reserved and the `kdump-tools` service will be enabled and active.
+Upon reboot, the `crashkernel` memory will be reserved and the `kdump-tools` service will be enabled and active.
 
 If you enable `kdump-tools` after a reboot, you will only need to issue the `kdump-config load` command to activate the `kdump` mechanism.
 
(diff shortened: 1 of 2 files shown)


docs/how-to/software/kernel-crash-dump.md
```diff
@@ -74,36 +53,156 @@ During the installation, you will be prompted with the following dialog:
...
+Which shows:
+
+```text
+[    0.004623] crashkernel reserved: 0x0000000060000000 - 0x0000000074000000 (320 MB)
+```
+
+This example output is for a system with 3GB memory which correctly maps to 320 MB.
+
+The defaults provided by the packaging try to be conservative to not waste too much.
+But in some rare cases that might break when the dump process fails to initialized.
```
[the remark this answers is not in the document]
s-makin: Valid suggestion

docs/how-to/software/kernel-crash-dump.md
```diff
@@ -74,36 +53,156 @@ During the installation, you will be prompted with the following dialog:
...
 
-As well, you will also need to edit `/etc/default/kdump-tools` to enable `kdump` by including the following line:
+## `kdump` is enabled by default where applicable
+
+On manual install the interactive question will define if it is enabled or not
+and therefore depend on the users choice, but in cases when it is automatically
+preinstalled there is a balance to strike.
+The tool wants to be helpful by default, because when the crash happens it is
+too late to enable crash dump functionality.
+But at the same time it needs to reserve memory for the crash kernel to operate
```
s-makin:
```suggestion
At the same time, it needs to reserve memory for the crash kernel to operate,
```

docs/how-to/software/kernel-crash-dump.md
```diff
@@ -74,36 +53,156 @@ During the installation, you will be prompted with the following dialog:
...
-If you're disabling `kdump-tools`, either set USE_KDUMP=0 or remove the line from the file.
+## Activating config changes
+
+If a reboot has not been done since installation/removal of the `kdump-tools` package,
+a reboot will be required in order to reconsider the `crashkernel= boot` parameter.
 
-If a reboot has not been done since installation of the `linux-crashdump` package, a reboot will be required in order to activate the `crashkernel= boot` parameter. Upon reboot, `kdump-tools` will be enabled and active.
+## Verify if kdump is ready
+
+Upon reboot, the crashkernel memory will be reserved and the `kdump-tools` service will be enabled and active.
```
s-makin:
```suggestion
Upon reboot, the `crashkernel` memory will be reserved and the `kdump-tools` service will be enabled and active.
```

docs/how-to/software/kernel-crash-dump.md
```diff
@@ -289,48 +370,149 @@ Verify that the *SysRQ* mechanism is enabled by looking at the value of the `/pr
...
 ```
 
 Once completed, the system will reboot to its normal operational mode. You will then find the kernel crash dump file, and related subdirectories, in the `/var/crash` directory by running, e.g. `ls /var/crash`, which produces the following:
 
 ```bash
-201809240744  kexec_cmd  linux-image-4.15.0-34-generic-201809240744.crash
+ll /var/crash
+```
+
+Showing
```
s-makin:
```suggestion
Showing:
```

docs/how-to/software/kernel-crash-dump.md
```diff
@@ -289,48 +370,149 @@ Verify that the *SysRQ* mechanism is enabled by looking at the value of the `/pr
...
+be copied away for analysis. But the examples are simplified to use e.g.
+`uname -r` to automatically select the right kernel - when running on a
+different system these might need to be adapted.
+```
+
+Then install the debug symbols for your kernel, assuming it is the same that
+runs currently that can be done with:
+
+```bash
+apt-get install linux-image-uname -r-dbgsym
```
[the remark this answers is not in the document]
cpaelzer: wow, good catch - it is correct and some copy and paste stripped it

PR #568, opened 2026-01-21
Update munin installation instructions

athos-ribeiro:
* Include directives to allow an external client to access the munin reports served by apache2
* Add note informing that the munin-plugins-extra package may already be installed

### DIFF
--- a/docs/how-to/observability/install-munin.md
+++ b/docs/how-to/observability/install-munin.md
@@ -45,7 +45,7 @@ On `server01` edit the `/etc/munin/munin.conf` file, adding the IP address for `
 Replace `server02` and `172.18.100.101` with the actual hostname and IP address for your server.
 ```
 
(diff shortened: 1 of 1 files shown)


docs/how-to/observability/install-munin.md
```diff
@@ -45,6 +45,16 @@ On `server01` edit the `/etc/munin/munin.conf` file, adding the IP address for `
 Replace `server02` and `172.18.100.101` with the actual hostname and IP address for your server.
 ```
 
+Then, if you wish to access the munin reports from a host other than `server01`, you will also need to edit the `Directory` directives in `/etc/apache2/conf-enabled/munin.conf`. For instance, you may want to allow access to some specific IP address. To do so, add the following snippet below each `Require local` entry:
```
s-makin:
```suggestion
Then, if you wish to access the Munin reports from a host other than `server01`, you will also need to edit the `Directory` directives in `/etc/apache2/conf-enabled/munin.conf`. For instance, you may want to allow access to some specific IP address. To do so, add the following snippet below each `Require local` entry:
```

PR #569, opened 2026-01-21
User management improvements

guilherme-puida:
### Description

Minor corrections related to changes in behavior.

An alternative to `w | grep <username> && pkill -f pts/<something>` is to use `loginctl terminate-user <username>` instead, but I'm not sure if both behave the same way in all situations.

---

### Checklist


s-makin:
> As a user, I would prefer to know exactly what version changes the behavior, even if it is not an LTS release.
> 
> I don't have permissions to merge, though :^)

I agree - disregarding the suggestion then.

Also that's weird. @cpaelzer will probably need to add you to the team.

PR #571, opened 2026-01-21
Improve clarity

Boroditsky:
### Description

Improves clarity and consistency of wording. Updates section introductions.


docs/explanation/networking.md
```diff
@@ -33,7 +33,7 @@ Configuring networks <networking/configuring-networks>
 
 ## Network tools
 
-In our how-to section we show to set up virtual switching using Open vSwitch (OVS) and the Data Plane Development Kit (DPDK) library. This page discusses DPDK in more detail.
+In our how-to section, we show how to set up virtual switching using Open vSwitch (OVS) and the Data Plane Development Kit (DPDK). This page discusses DPDK in more detail.
```
s-makin:
```suggestion
In our how-to section, we show how to set up virtual switching using Open vSwitch (OVS) and the Data Plane Development Kit (DPDK) library. This page discusses DPDK in more detail.
```

docs/explanation/networking.md
```diff
@@ -67,9 +67,9 @@ Time synchronisation <networking/about-time-synchronisation>
 
 ## Network shares
 
-Samba and Active Directory are two tools you will need if you want to share resources and directories between Linux and Windows systems.
+Two tools commonly used to share resources and directories between Linux and Windows systems are Samba and Active Directory.
```
s-makin: We prefer to avoid use of passive language wherever possible. In addition, the change to "commonly used" changes the meaning of the sentence. Let's revert this change.

docs/explanation/software.md
```diff
@@ -11,7 +11,9 @@ This section explains how software is managed on Ubuntu Server and outlines reco
 
 ## Package sources
 
-* {ref}`Third party repository usage <third-party-repository-usage>` gives some best practices and guidance if you need to use Third Party software
+Ubuntu Server software is typically installed from official repositories, with optional support for third-party sources.
```
s-makin: This implies something that I think is not correct, i.e. that we support software from third part sources when we don't. Can we remove this sentence?

s-makin (changes requested):
Thanks for taking the time to contribute, @Boroditsky. I've left some suggestions in-line to help get it ready to land.

By way of explanation, some of the proposed changes are quite minor and don't necessarily improve the clarity of what was there, e.g. changing one word for a longer synonym, or switching from active voice to the passive voice (which is harder to read/parse). A word change should make the text more precise/accurate/understandable - if it only changes the flavour or is a personal preference, we don't take those.

The rest of the changes are welcome, with only a few minor nitpicks. I hope the distinction will be clear from the comments below, please let me know if you'd like me to clarify anything :)

Boroditsky:
Thanks for the ping @s-makin — you’re right, the PR was still pointing to my fork’s main. I’ve now force-pushed the updated commit to main, so the changes should be visible.

PR #683, opened 2026-03-30
openvpn: document usage of tls-crypt instead of tls-auth

guilherme-puida:
### Description

Using `tls-crypt` provides encryption to the key exchange phase before the TLS session is established.

I tested this by creating two Resolute LXD containers and configuring one as client and one as server as described in the guide. Then, I switched to `tls-crypt` and verified that the connection was still working. Unsure if any more testing is needed.

I'll also verify that the output of systemd commands are up-to-date, so I'm opening this PR as draft for now.

Do note that this option was introduced in OpenVPN 2.4.0, which is present in Ubuntu releases >= Bionic. Do we need to keep the old `tls-auth` configuration documented somewhere? Xenial is already EOL, and ESM is ending very soon.

### Checklist

### DIFF
--- a/docs/how-to/security/install-openvpn.md
+++ b/docs/how-to/security/install-openvpn.md
@@ -115,6 +115,12 @@ dh dh.pem
 tls-crypt ta.key
 ```
 
+:::{note}
+The `tls-crypt` option was introduced in OpenVPN 2.4.0. If you are running an
+Ubuntu Server release older than 18.04 LTS (Bionic Beaver), please use
+`tls-auth ta.key 0` instead.
+:::
+
 Complete this set with a TLS Authentication (TA) key in `/etc/openvpn` for `tls-crypt` like this:
 
 ```bash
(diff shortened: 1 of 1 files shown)


s-makin (approved):
lgtm :)

s-makin:
> Do note that this option was introduced in OpenVPN 2.4.0, which is present in Ubuntu releases >= Bionic. Do we need to keep the old `tls-auth` configuration documented somewhere? Xenial is already EOL, and ESM is ending very soon.

In terms of keeping references for older versions of Ubuntu, a note block that points out the switch (and when it crossed over) would be sufficient. I'll hold off on merging in case you want to add something like that. I don't consider it a blocker though.

guilherme-puida:
Thanks, Sally! I decided to add a note anyways :)

s-makin:
> Thanks, Sally! I decided to add a note anyways :)

Looks good!

PR #689, opened 2026-04-09
Update homepage to match new standard

s-makin:
### Description

This cycle, we were requested to experiment with methods for adopting the new [homepage standard model](https://docs.google.com/document/d/1Zw-UoQzHMSQjATLXjjvd9GKu5qU2mHmqxtKck8nGkbU/edit?tab=t.h96gm4u3zba1) from the design team. 

The Server documentation is something of an outlier to the rest of the docs that have so far adopted it, in the sense that it is not a single product, and we have 250 pages to wrestle with, which adds some challenges and necessary constraints.

However, this PR represents my first attempt to adhere to the model in a way that makes sense for the Server documentation. Please share your opinions! I want to know what you do/don't like, where you think it could be improved, etc.

---

### Checklist

---

### AI disclosure

Since the home page standard model is well described and expected to be an artifact (similar to a style guide) that we must comply with, it seemed like a good candidate for an instruction file. I have therefore made this my first test of using instructions.

The initial iteration was done by Claude Sonnet 4.6 based on my prompt to compare the homepage against the instruction file. It did a good job of identifying where the homepage did not comply with the standard model, and provided me with a reasonable starting point which I then iterated over several times to reorganize and rework. I then asked Claude to update the instructions based on my heavy editing. I have not included the instruction file as it does not comply with the actual standard model, but merely represents my interpretation of it (which will be subject to change).

I intend to add an instructions folder to the repo with all of the standards we need to comply with so that PRs can be checked against them. I'll be doing a similar thing with agents and skills, but instructions are first.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -6,85 +6,133 @@ myst:
 
 # Ubuntu Server documentation
 
-```{only} latex
-**Overview**
-```
+**Ubuntu Server** is a version of the Ubuntu operating system designed and engineered as a backbone for the internet.
 
-**Ubuntu Server** is a version of the Ubuntu operating system designed and
-engineered as a backbone for the internet.
+Ubuntu Server brings economic and technical scalability to your data center, public or private. Whether you want to deploy an OpenStack cloud, a Kubernetes cluster or a 50,000-node render farm, Ubuntu Server delivers the best value scale-out performance available.
 
-Ubuntu Server brings economic and technical scalability to your data center,
-public or private. Whether you want to deploy an OpenStack cloud, a Kubernetes
-cluster or a 50,000-node render farm, Ubuntu Server delivers the best value
-scale-out performance available.
+This documentation always targets the latest LTS version of Ubuntu. When there are differences towards older LTS releases, they are noted in the guide.
 
-This documentation always targets the latest LTS version of Ubuntu. When
-there are differences towards older LTS releases, they are noted in the
-guide.
-
-```{only} html
 ## In this documentation
-```
 
-```{only} latex
-**In this documentation**
-```
+### Getting started
 
-::::{grid} 1 1 2 2
-:gutter: 3
-:margin: 0
+The tutorial walks you through a fresh installation and the first steps of operating an Ubuntu Server.
 
-:::{grid-item-card} {ref}`Tutorial <tutorial>`
-Get started - a hands-on introduction to Ubuntu Server for new users
-:::
+* **Installation**: {ref}`System requirements <system-requirements>` • {ref}`Basic installation <basic-installation>` • {ref}`Ubuntu Pro subscription <attach-your-ubuntu-pro-subscription>`
+* **System basics**: {ref}`Managing software <managing-software>` • {ref}`Customizing package files <changing-package-files>` • {ref}`third-party-repository-usage`
 
-:::{grid-item-card} {ref}`How-to guides <how-to>`
-Step-by-step guides covering key operations and common tasks
-:::
 
-:::{grid-item-card} {ref}`Explanation <explanation>`
-Concepts -
-Overviews of key topics, as well as more in-depth discussion and
-clarification 
-:::
+### Networking
 
-:::{grid-item-card} {ref}`Reference <reference>`
-Technical information -
-{ref}`system requirements <system-requirements>`,
-package specifications and architecture
-:::
-::::
+Configure network interfaces, provide name and address services, share resources, and synchronize time.
 
-```{only} html
+* **Understanding networks**: {ref}`Key concepts <networking-key-concepts>` • {ref}`About Netplan <about-netplan>` • {ref}`Time synchronization <about-time-synchronisation>`
+* **Configuration**: {ref}`Using Netplan <configuring-networks>` • {ref}`DNS <install-dns>` • {ref}`DHCP <install-isc-kea>`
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -6,85 +6,133 @@ myst:
...
+
+
+### Security
+
+Protect your system with access controls, firewalls, encrypted communications, and network authentication.
+
+* **Understanding security**: {ref}`Introduction <introduction-to-security>` • {ref}`security-suggestions` • {ref}`DNS Security Extensions (DNSSEC) <dnssec>`
+* **Basic security**: {ref}`User management <user-management>` • {ref}`Firewalls <firewalls>` • {ref}`AppArmor <apparmor>`
+* **Authentication**: {ref}`Install Kerberos <install-a-kerberos-server>` • {ref}`Set up SSSD with LDAP <sssd-with-ldap>` • {ref}`Smart cards <smart-card-authentication>`
+* **Cryptography**: {ref}`Cryptographic libraries <introduction-to-crypto-libraries>` • {ref}`Install OpenSSH <openssh-server>`
```
guilherme-puida: I feel like OpenSSH would make more sense in the networking section.

guilherme-puida (approved):
Inline comment is just a quick thought. +1 on this even without any changes :)

docs/index.md
s-makin: What makes networking a better fit in your opinion?

(I'm happy to add some of the crypto library pages we have instead, just want to understand the reasoning)

guilherme-puida: Hm, it's hard to articulate but I think of "remote login/access" before "ssh secure protocol" when OpenSSH is mentioned.

lvoytek (approved):
LGTM too! I think the organization of topics is reasonable

cpaelzer (changes requested):
+1 on starting with getting started and its content.
This is where the new tutorials on basic use out of cloudbase would be added right?

ok on the networking section, it feels a bit weak though.
Could you please add https://canonical-ubuntu-server-documentation--689.com.readthedocs.build/explanation/networking/about-dpdk/ and https://canonical-ubuntu-server-documentation--689.com.readthedocs.build/how-to/networking/dpdk-with-open-vswitch/ there ?

security - ok'ish
But I muss here right at the start also a link to https://documentation.ubuntu.com/security/
Also almost all sub-pages that link to https://ubuntu.com/security (product) like the introduction should instead link to the security content https://documentation.ubuntu.com/security/ please

+2 I love the grouped Managing your system

Can you for databases please file an issue to document valkey a bit at some point?
Not a problem of this PR but became apparent.

etckeeper from the line "version control" you have in "Data and Storage" should move to "Managing your system" please.

I've seen you not mention all pages up here, can we remove "Web programming: PHP • Ruby on Rails" here. The content is fine, but those are very old stacks. They are used for sure but having "only" them makes us a dinosaur. Not mentioning them in the overview seems better.

"Understanding virtualization: Intro to virtualisation • VM tools • Container tools"
Of this "Container tools" needs to go to the containers line.

s-makin:
> +1 on starting with getting started and its content. This is where the new tutorials on basic use out of cloudbase would be added right?

Yes, I'd be putting them here :)

> ok on the networking section, it feels a bit weak though. Could you please add [snip]

I definitely can, thanks for the suggestion!
 
> security - ok'ish But I muss here right at the start also a link to https://documentation.ubuntu.com/security/ 

I am not yet sure if this is something allowed or not by the model (i.e. how strict it is), have brought this up as a discussion point among the other TAs to clarify.

> Almost all sub-pages that link to u.c/security (product) like the introduction should instead link to the security content https://docs.u.c/security/ please

> +1, will do!

> Can you for databases please file an issue to document valkey a bit at some point? Not a problem of this PR but became apparent.

This was filed during the docsprint, and will be tackled by Kevin (with Lena's help) next cycle. That and redis can both be added to this section then.

> etckeeper from the line "version control" you have in "Data and Storage" should move to "Managing your system" please.

Ack, will do

> I've seen you not mention all pages up here, can we remove "Web programming: PHP • Ruby on Rails" here.

Will do this too :)

> "Understanding virtualization: Intro to virtualisation • VM tools • Container tools" Of this "Container tools" needs to go to the containers line.

Good point, although then it breaks my self-imposed rule of 3 links per line. I think I'll just drop it - we don't have so many containers to document like we do in VMs where I think the disambiguation is more needed.

docs/index.md
s-makin: Ack, that makes sense, and tbh is more aligned with what we document in that page. Thanks for the suggestion! I'll make that change.

s-makin: Ended up adding a "securing your network" row, with OpenSSH, Firewalls, and DNSSEC (felt weird to have it there by itself)

s-makin:
I'm done applying review feedback now and ready for round 2. Most suggestions were applied as-is.

* After some discussion among TAs, I've decided to include the link to the security docs on the front page even though it's not "in this documentation". Security is such a hot topic, the Security docs contain a lot of info relevant to a Server sysadmin, and we want the Security docs to be easily discoverable, so it makes sense imo to include it.
* I couldn't resolve the "networking or security" question in a way that made sense, so I've added not just OpenSSH to networking, but added a row for "securing your network" that also includes the pages with the same overlap.
* Added certificates and openssl to the crypto line under security. After moving openssh out it was a bit "light" and we have plenty of pages under crypto that it makes sense to include, so I went with the most popular ones.

cpaelzer (approved):
Thank you for the adaptations - +1 now

PR #697, opened 2026-04-23
Tutorial pt 1 - welcome to the terminal

s-makin:
### Description

Part 1 of aligning our content with our external training partners. This tutorial covers the same basic terminal instructions, where part 2 will go into more depth.

Note I have also added the `sphinx-terminal` extension, I'm interested to hear feedback about whether people like this or not.

---

### Checklist

---

---


docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+* **Knowledge:**
+
+  None! You don't even need to use an Ubuntu machine -- we will use Multipass to create a virtual Ubuntu environment to play with.
+
+* **Hardware:**
+
+  The Multipass virtual machine needs at least **5 GB of disk space**, and **1 GB of memory**.
+
+
+## Install Multipass
```
grayson-wolf, line 22: Should we be using multipass in the tutorial for people already on Linux? Having linux users spin up the VM through LXD might be a better choice, though it is also simpler to just have a unified tutorial for all, so I don't have particularly strong feelings on this, just raising it for consideration.

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+:copy:
+:user: ubuntu
+:host: tutorial
+:dir: ~/my-folder
+ls
+
+my-first-file.txt my-second-file.txt my-third-file.txt
+```
+
+Did you notice how the shell prompt changed to `ubuntu@tutorial:~/my-folder$` after we used the `cd` command? This is our biggest navigation help! The tilde (`~`) symbol signifies our home directory, which was our starting point. This is often called the "root directory". All the directories and subdirectories we create can trace a path back to the root directory.
```
grayson-wolf: right now this reads to me as saying that `~` is called the root directory,  while most documentation will usually just call this `~` or "your home directory" or `$HOME` and explicitly only use "root" to refer to the entire filesystem root `/`. 

I do think mentioning how the entire filesystem is all one big tree at around this point in the docs is good, so maybe add a bit explicitly explaining that `~` actually is short for `/home/your_user_name` and that the _actual_ root from which everything descends is `/`. Especially since later in this tutorial we see `/home/ubuntu` in place of `~` in things like the output `pwd` and get a fuller explanation of the whole structure later.

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+
+```{terminal}
+:copy:
+:user: ubuntu
+:host: tutorial
+:dir: ~/my-folder
+mkdir second-level
+```
+
+Another helpful feature of the command line is "tab completion". If a command, directory, or file already exists, we can start typing it and then press the {kbd}`TAB` key on the keyboard to complete it. Let's navigate to the directory we just made by typing `cd s`, and then press {kbd}`TAB`. The CLI automatically fills in the rest!
```
grayson-wolf:
```suggestion
Another helpful feature of the command line is "tab completion". If a command, directory, or file already exists, we can start typing it and then press the {kbd}`TAB` key on the keyboard to complete it. Let's navigate to the directory we just made by typing `cd s`, and then press {kbd}`TAB`. The shell automatically fills in the rest!
```

nitpick

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+:user: ubuntu
+:host: tutorial
+:dir: ~
+ls
+
+file1.txt file2.txt new-folder
+```
+
+So what if we want more copies of the same file, but in our `new-folder`?
+For that, we can use `cp` again, but give the directory as the destination:
```
grayson-wolf, line 556: maybe mention that the terminal will not check for name collisions and just gleefully overwrite files here?

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+drwxrwxr-x 2 ubuntu ubuntu 4096 Apr 21 18:00 new-folder
+drwxrwxr-x 2 ubuntu ubuntu 4096 Apr 21 18:00 new-folder-backup
+```
+
+The output looks quite complicated at first, so let's break it down using the first row of the output.
+
+| Column         | What it is  | Description |
+| ---            | ---         | --- |
+| `-rw-rw-r--`   | Permissions | The type of file and who can read/write/execute it |
+| `1`            | Links       | The number of hard links to the file or directory |
```
grayson-wolf:
```suggestion
| `1`            | Links       | How many distinct filesystem paths refer to this file or directory (these are called "hard links") |
```

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+From this we can see that `755` means: the owner has read, write and execute (7) privileges, the group has read and execute (5), while all others have read and execute (5).
+
+Some common permission combinations:
+
+| Number | Symbolic | Meaning |
+|--------|----------|---------|
+| `644` | `rw-r--r--` | Owner can read/write; others can only read. Suitable for most files. |
+| `755` | `rwxr-xr-x` | Owner can do everything; others can read and execute. Suitable for scripts and directories. |
+| `700` | `rwx------` | Only the owner has any access. |
+| `600` | `rw-------` | Owner can read/write; no one else has any access. |
```
grayson-wolf, line 791:
```suggestion
| `600` | `rw-------` | Owner can read/write; no one else has any access. |
| `777` | `rwxrwxrwx` | Full access to everyone. **This is a dangerous permission to grant**, but might occasionally be useful for debugging permission-related problems or on temporary files. |
```

People really shouldn't use 777 unless they know what they're doing, but I think it deserves a mention if only to say "please don't use this one unless you know what you're doing, and if you see it you should probably change it"

docs/reference/cli-cheatsheet.md
grayson-wolf:
```suggestion
| `rm <file>` | Remove a file |
| `rm -r <directory>` | Remove a non-empty directory and all of its contents |
```

`rm -r` warrants an explicit mention here IMO

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+
+bin   dev   home  lib64   media  opt   root  sbin  srv  tmp  usr  var
+boot  etc   lib   mnt    proc  run   snap  sys
+```
+
+Here is what the most important directories contain:
+
+| Directory | Purpose |
+|-----------|---------|
+| `/bin` and `/sbin` | Essential command binaries (e.g., `ls`, `cp`, `mv`) and system binaries (e.g. `iptables`, `reboot`) respectively |
```
grayson-wolf:
```suggestion
| `/bin` | Essential command binaries (e.g. `ls`, `cp`, `mv`)
| `/sbin` | System binaries (e.g `iptables`, `reboot`)
```

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+| `/boot` | Static files needed to boot the system, including the Linux kernel and bootloader files |
+| `/dev` | Device files (e.g. `/dev/sda` and `/dev/null`) |
+| `/etc` | System configuration files (e.g. `/etc/group` and `/etc/password`) |
+| `/home` | User home directories (optional, but standard) |
+| `/lib` | Essential shared libraries (needed by the binaries in `/bin` and `/sbin`) |
+| `/media` and `/mnt` | Temporary mount points for removable media (like USBs and CDs) and for manually mounting filesystems (respectively) |
```
grayson-wolf:
```suggestion 
| `/media` | Mount points for temporarily mounted removable media (like USBs or CDs)
| `/mnt` | Mount points for manually mounted filesystems (kept seperate to avoid clashes)
```

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+Where: 
+* `group_name` is the human-readable name of the group (what you see when you run `ls -l`)
+* `password` is always shown as `x` in modern Linux systems - the actual (encrypted) group password is stored in the secure `/etc/passwd` file.
+* `GID` is the group ID, which is a unique numerical integer assigned to the group.
+* `user_list` is a comma-separated list of usernames that are members of the group.
+
+
+## The Filesystem Hierarchy Standard
+
+Now that we know how to move around the filesystem, it's worth understanding how it is organised. Linux follows the **Filesystem Hierarchy Standard** (FHS), which defines where different types of files and directories belong. This is maintained by the Linux Foundation, and it means the structure of any Linux system is predictable and consistent.
```
grayson-wolf:
```suggestion
Now that we know how to move around the filesystem, it's worth understanding how it is organised. Most Linux distributions, including Ubuntu, follow the **Filesystem Hierarchy Standard** (FHS), which defines where different types of files and directories belong. This is maintained by the Linux Foundation, and it means the structure of any conforming system is predictable and consistent.
```

Maximal pedantry: Not _every_ Linux system is FHS compliant (hi, NixOS!)

docs/reference/cli-cheatsheet.md
grayson-wolf:
```suggestion
| `ls -la` | Long listing including hidden files (also usable with `ll`) |
```
i think this extremely useful alias is worth a mention

cpaelzer, line 15: depending on how entry or super-entry friendly you want to make those I could think of prepending the common question answered by the tool here. Like under `## Naviagation` having a `### where am I?` => pwd, `### what files are here?` => ls, `### How can I go elsewhere?` => cd.

cpaelzer: I think this needs examples to differ rename and move

rename `mv <oldfilename> <newfilename>`
move `mv <oldplace>/file <newplace>/file`

WDYT?

cpaelzer: you say "10" as it is the default - should we add `-n 10` to show them how to influence and mention in the text that it is the default? Unsure, maybe better keep the command simple  and mention (-n to control) in the test?

cpaelzer, line 86: Well, this is actually a good point - all of the above should be links to their man pages.
And since you can't link inside of `code` you should make the whole example a link each I think

docs/tutorial/images/shell.png
cpaelzer, line 1: I have seen this before, but it suggests that applications would go through the CLI tools to do their job.

I think this would be better if the outer two would be one.
Make them half yellow half green but on the same level.
They both speak to the kernel at equal levels.

Mark the passage of that layer to the kernel with "system calls"

And shell needs a counterpart as well it is IMHO more like
SHELL <-> GUI
CLI tools <-> vs graphical tools

And kernel to hardware system interfaces via various tihngs (e.g. mmio) but let us simplify this with "drivers"

Here a suggestion

Image

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+```
+
+We've just successfully run our first command on the CLI!
+
+
+### The shell
+
+You may be wondering: "why we need a shell at all? Why can't we just tell the operating system what to do directly?"
+
+This is a good question. In short, it's because the hardware in our machines doesn't speak "human", and it's difficult and time consuming for us to try and speak "computer". So, we use the shell as a translation layer to convert what we want into precise **system calls** that the computer can understand.
```
cpaelzer: I do not believe in this to be a good explanation.
You can very much run all the CLI tools from the GUI.
try ALT+F2 which wil open "run a command" and you can ask it to run any CLI tool.

The problem is, where is its input, where its ouput?

Yes - the diff is that those tools do not have graphical input or output, they are text based and mostly perform one general function. More complex tools extend on that. Here you might want to refer to the very famouts and basic philosophy of https://harmful.cat-v.org/cat-v/unix_prog_design.pdf which has this statement

And then in turn what a shell is about IMHO is:
a) basic: providing a way to provide input and consume output of CLI programs
b) efficiency: no graphical handling often makes the tools more efficient and for the same reason servers often do not even have any GUI installed as it is a burden not needed there for operating services
c) the ability to issue multiple of them in the same context e.g. working dir
d) provide the ability to connect multiple of them to more capable function (via arguments, control structures like if/loops, subshells to create arguments for the next and pipes to connect them). All this in (d) follows here later or much later.

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+
+### The shell
+
+You may be wondering: "why we need a shell at all? Why can't we just tell the operating system what to do directly?"
+
+This is a good question. In short, it's because the hardware in our machines doesn't speak "human", and it's difficult and time consuming for us to try and speak "computer". So, we use the shell as a translation layer to convert what we want into precise **system calls** that the computer can understand. 
+
+![shell](./images/shell.png)
+
+This makes computers much easier to work with, and it means we don't need to be fluent in binary code. Instead, we can learn commands, which are a sort of shorthand.
```
cpaelzer: Again, this isn't what it is.
It is more about the fact that calling tools within control structures and combining them allows even simple tools to quickly outperform what even complex GUI applications can do. Think of the CLI as a set of building blocks (lego) while the GUI mostly has predetermined use cases (playmobil). If you are withint the predetermined use cases the GUI might be more comfortable but the CLI quickly gets more ppowerful as you can make it do what you need.

Example
- rename a file
  - GUI open file manager, right click rename
  - CLI `mv oldname newname`
  - they are about equal
- rename all files ending in .txt to prepend current time of day
  - GUI a lot of clicking and typing and more
  - `rename "s/^/$(date +%H%M)-/" *.txt` pulling in the date automatically, affecting all that match a patern and done

So much for a better explanation where they differ.
If you can find a non trademark example of lego/playmobil that is actually a good comparison

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+
+In fact, Multipass *itself* is a shell. Instead of being a translation layer around our physical machine, it's a translation layer around the *virtual* machine we've just made. Let's access our virtual machine now by running this command:
```
cpaelzer: "In fact, Multipass *itself* is a shell" - umm, very no :-)

Just drop the whole line I think.

cpaelzer: Actually I now found what you want to say ...

Suggestion:

The virtual machine provided by multipass also comes without the burden of a GUI and hence the easy, general access it provides you to run commands on the shell is abstracting all the details of virtualization and allowing you to open a shell in running instance.

(and then follows the command multipass shell tutorial)

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+To run a command as administrator (user "root"), use "sudo <command>".
+See "man sudo_root" for details.
+```
+
+You've probably noticed that your shell prompt has changed, and now says `ubuntu@tutorial:~$` -- `tutorial` is the name of our virtual machine, and `ubuntu` is the default username in a Multipass virtual machine. This is how we know we're now playing in our virtual sandbox, and not in our physical computer.
+
+
+### Other ways to use commands
+
+As we've now seen, commands are the instructions that we give the computer by typing them into the terminal. In our documentation, commands are `formatted like this`.
```
cpaelzer:
```suggestion
As we've now seen, CLI commands are calls to start executable binaries that we give the computer by typing them into the terminal. In our documentation, commands are `formatted like this`.
```

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+
+Sometimes, we might want to run a whole set of commands at once.
+For convenience, we can save them in a file called a **shell script**, in the order we want the computer to run them. Then, we just need to command the computer to run the script, and it runs the commands in the script by itself.
```
cpaelzer:
```suggestion
For convenience, we can save them in a file called a **shell script**, in the order we want the shell to execute them. Then, we just need to tell the shell to run the script, and it runs the commands in the script by itself.
```

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+
+As you become more familiar with the command line, you can also begin to define your own commands built from other commands. These are called **aliases**, and they can be very handy shortcuts for small sets of commands you want to run often.
+
+
+## Navigating the filesystem
```
cpaelzer, line 186: I now see the commands of the cheat sheet coming.
Should we link to it somewhere here to say "here are the details but you might use ... for later reference"?

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+:copy:
+:user: ubuntu
+:host: tutorial
+:dir: ~
+ls
+
+my-first-file.txt my-second-file.txt my-third-file.txt
+```
+
+We can also create folders to help keep all our files organized using the `mkdir` command:
```
cpaelzer, line 224: Globally - all those `commandname` should be ``` [`commandname`](link-to-manpage) ```

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+
+The `mkdir` command is short for "make directory". A directory is exactly the same as a folder, but it abbreviates to the conveniently short `dir` so it's used more often. We'll use them interchangeably.
+
+Let's now move the three files we created into `my-folder`, so we can keep our current folder nice and tidy. We can do this using the `mv` command.
+
+Since it doesn't know which files to move, we must give some **arguments** to the move (`mv`) command. Actually, we already did this with our `touch` and `mkdir` commands earlier:
+
+```bash
+touch FILE(S)
+touch <file-name> # Create one file
```
cpaelzer, line 256: Somewhere above you explained the generic style e.g. `cmd` for commands.
Since this is beginner level I think there you need to explain that `<foo>` is a placeholder for something they add. Because in reality this is redirecting input and output.

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+:dir: ~
+rm file1.txt
+```
+
+```{warning}
+Remember, there is no recycle bin in the Linux terminal -- once you `rm` a file, it is gone. On your real system, double-check what you are deleting before running this command.
+```
+
+
+## Understanding file types
```
cpaelzer:
```suggestion
## Understanding attributes of a file
```

a file type is something else like is it a graphic or sound file

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -0,0 +1,1089 @@
...
+:user: ubuntu
+:host: tutorial
+:dir: ~
+man -k copy
+```
+
+This lists every command whose description mentions "copy", which can help us find the right tool for the job.
+
+
+## Next steps
```
cpaelzer, line 1082: Will there be a second one which includes the next obvious things (if not could it go here please)?
- redirecting output and input `<` and `>`
- using pipes to connect stdin/stdout
- starting things in background, + `&`, `bg`/`fg`
- use of sub-shells `cmd1 $(cmd2 | xargs)` 
- advances redirecting output and input `<<< heredocs` and `2>&1 >/dev/null` and such

cpaelzer (changes requested):
Great, still with so much content there had to be some feedback :-)
I'm mostly against some comparisons and examples in the intro to not set up wrong assumptions, then a few hints how to do better and the final question when the next things come (here or extra)

grayson-wolf (changes requested):

docs/reference/cli-cheatsheet.md
s-makin: The intended audience for this is not necessarily the super-beginner entry level, but rather people who just want to look stuff up quickly for work - the expanded version of the 1-pager PDF cheatsheet Oliver created for Noble. Since the tutorial page will cover the super-beginner entry level intro to the commands, I think it's fine for this to stand as a more general reference page.

s-makin: TIL :D

s-makin: Good idea, added

s-makin: Good call, added

s-makin: I didn't know you could do that, I'll add it

s-makin: In that case, do you think it would be more helpful to also put the getting help section at the top?

cpaelzer: ok then - resolving this one

cpaelzer: no, I'd keep the help section at the bottom

s-makin: This turns out not to work very well, unless I use hardcoded links (which I don't want to do, otherwise we'll have to update them every 6 months). I'm adding specific manpage links for each to the description instead.

docs/tutorial/welcome-to-the-terminal.md
s-makin: This tutorial is aimed at people who may have (potentially) never used either Linux or a virtual machine before. Multipass is arguably the easiest way for a complete beginner to access Ubuntu, particularly if they're coming from windows or mac. I'd rather have the same instruction for all three (rather than using LXD for Linux users, WSL for windows and Multipass for mac) just to ensure that the results the user sees are the same and predictable.

We're starting to get more tutorial content, so one option is if I make all the tutorials into a series, then we could split for "beginners" (multipass) and "intermediate" (LXD) and "advanced" (QEMU). WDYT?

docs/tutorial/welcome-to-the-terminal.md
s-makin: That's all in part 2 :)

docs/tutorial/welcome-to-the-terminal.md
s-makin: Thanks - I think I oversimplified in the wrong direction.

docs/tutorial/welcome-to-the-terminal.md
s-makin: I can't add the term "executable binaries" without explaining it. Do we want to add that level of detail in such a beginner level tutorial?

docs/tutorial/welcome-to-the-terminal.md
s-makin: Done

docs/tutorial/welcome-to-the-terminal.md
s-makin: I probably won't add that here - otherwise we're distracting people while they're on the path we're setting out for them. I do agree we should have that link though, so I'll add it at the end of the page.

docs/tutorial/welcome-to-the-terminal.md
s-makin: I'd prefer not to do that in this tutorial - adding too many links in a tutorial environment can be a significant distraction at the best of times, but here we want to encourage them to play around in the CLI and that's why where I've suggested looking at manpages, that they use `man <cmd>` instead (and only after the command has been introduced and they've run it). I added manpage links to the reference cheatsheet, where I agree that it makes sense.

docs/tutorial/welcome-to-the-terminal.md
s-makin: That's a great point. I've never been a fan of the "foo" "bar" paradigm, but it's so ubiquitous I probably ought to mention it.

s-makin: Added

docs/tutorial/welcome-to-the-terminal.md
s-makin: Done!

docs/tutorial/welcome-to-the-terminal.md
s-makin: Good idea, also including "gleefully" because I like anthropomorphising machines

docs/tutorial/welcome-to-the-terminal.md
s-makin: adopted, you're right - I should have added a definition for that

docs/tutorial/welcome-to-the-terminal.md
s-makin: Nice, added!

docs/tutorial/welcome-to-the-terminal.md
s-makin: I'm not averse to a little light pedantry :D added

(also, I never played with Nix so I didn't know that, thanks for the info!)

docs/tutorial/welcome-to-the-terminal.md
s-makin: Sometimes I'm a bit overzealous in my desire to tidy things up lol

docs/tutorial/welcome-to-the-terminal.md
s-makin: Added this too, thanks

docs/tutorial/welcome-to-the-terminal.md
s-makin: Revised all three paragraphs, lmk if you think they're better

docs/tutorial/welcome-to-the-terminal.md
s-makin: Revised this

docs/tutorial/welcome-to-the-terminal.md
s-makin: I think I might just defer discussions about root until after the FHS section (probably in tutorial part 2). On reflection, using root in the colloquial way I hear and see people using it is probably the wrong decision for a tutorial this basic. I've added the clarifications you suggested, but removed references to root here.

docs/tutorial/images/shell.png
s-makin: I tried doing this in Mermaid but it wasn't working as well as I'd hoped, so worked it up in Inkscape instead. Lmk if I'm on the right lines.

docs/tutorial/welcome-to-the-terminal.md
grayson-wolf: Seems like a good idea, especially since once you actually get into the Linux container (wherever it may be) the tutorials are identical (since I don't think any tutorial level content will care if its virtualized). For now I think multipass alone is good, so I'm resolving this

grayson-wolf:
Just looked over the newest changes, LGTM once any leftover changes of Christian's are resolved.

s-makin:
Double checked all outstanding comments - all resolved one way or another. We can continue to iterate after merging :)

PR #702, opened 2026-04-30
Replace icons with text in table

### LINKED ISSUE #700: Acrobat Reader doesn't render special signs in table
### Page URL

https://ubuntu.com/server/docs/_/downloads/en/latest/pdf/

### Select an option

- [x] I encountered a technical issue (e.g., broken link, image not loading)

### Issue Description

It seems Acrobat Reader is unable to properly render the information in the table on p.831 of 981 of the Ubuntu Server documentation found at https://ubuntu.com/server/docs/_/downloads/en/latest/pdf/ :

Image

I am trying to read this on a PC with Win11, Acrobat Reader Version 2026.001.21483 (latest version as of 28 April 2026, no further updates available).

Maybe using something plain like “Y”, “N”, “O”, “X”, or whatever makes most sense, rather than exotic symbols would be better?

s-makin:
### Description

As reported in #700, the table is not rendering correctly in the PDF due to a lack of encoding for those symbols. This PR replaces the emoji symbols with plain text.

---

### Related Issue

- Fixes: #700 

### Checklist

### DIFF
--- a/docs/explanation/active-directory/choosing-an-integration-method.md
+++ b/docs/explanation/active-directory/choosing-an-integration-method.md
@@ -14,16 +14,16 @@ Three main criteria need to be evaluated:
 * member server or workstation: Is the Ubuntu system that is being joined to Active Directory going to be a member server (it provides authentication and shares), or just a workstation/desktop which will not share files via the SMB protocol? If a workstation is also going to share files via SMB, then consider it a member server.
 * Deterministic Linux ID: is it important that the same Active Directory user obtains the same Linux ID on all joined systems? For example, if NFS exports will be used between the joined systems.
 
-|Requirement / Method      |SSSD |idmap\_rid  |idmap\_autorid|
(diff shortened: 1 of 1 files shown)


s-makin:
PDF isn't built in the RTD preview, need to merge to check how this fix renders.

renanrodrigo:
@s-makin when should we expect the pdf to be updated? I double checked and it's the same for me

s-makin:
> @s-makin when should we expect the pdf to be updated? I double checked and it's the same for me

It takes a long time to build, and then the cache is cleared every few hours (I think). However, there's quite a lot of issues in the LaTeX build log that could be contributing to this, mostly around characters being missing in the Ubuntu fonts. It's possible that switching to an alternative font that includes all the characters could resolve those, but I need to do some investigation. I've got the PDF building locally right now, so I'll see if I can identify what the real problem is.

PR #703, opened 2026-05-14
Move images to repo

s-makin:
### Description

This week's linkcheck revealed that we still have some images hosted on the asset manager instead of inside the repository where the docs live.

The current guidance is that all images should live alongside the docs, so in this PR I move all the docs in the asset manager to the repo, and update the paths to each file accordingly.

The next time there is an outage on ubuntu.com sites, the linkchecker should no longer be triggered by images being in the wrong location. 

AI disclosure: I used Claude Sonnet to speed up the fetching and reorganising of the images, and to validate the paths against each image.

### Checklist

- [ ] I have tested my changes, and they work as expected.


s-makin:
I've validated the pages changed to ensure continuity and that no errors have been introduced - all good.

PR #711, opened 2026-05-26
openldap: add guide for simple ldap client using nslcd

### LINKED ISSUE #585: openldap improvements
### install section

- meta description: remove "set up database suffix"
- the first command is `sudo apt install slapd ldap-utils`, which will prompt for an administrator password. we should mention this, and advise what to enter there (or what it means when this is left empty)
-  "Change the instance suffix":
  - first explain what the suffix is (briefly) -> "the top-most tree entry where we'll create entries below" (the base dn tree entry)
  - make the relation between the displayed domain input box and that suffix more clear (just by reordering the sentences)
  - the prompt also continues to ask for the organization name (which will be attribute `o=` (for `objectClass: organization`) for the base dn tree entry)
  - and the adminitrator password (again) -> mention this is the password for `cn=admin,dc=example,dc=com` then.
  - it asks to move away the previous database to `/var/backups/`
- "Who am I?"
  - since this already talks about `peercred` and socket authentication, which we'll set up in a basic way in "populate the directory", we should add a forward reference here saying "we can use this to allow the root user to do any ldap tree modifications, see next ..."
- "Populate the directory"
  - before "Add the content:" reference to access control where add the description (see below, the 'access control' section) to set up passwordless authentication for writing to the example.com tree 
  - now we can instead just use `sudo ldapadd -H ldapi:/// -Y EXTERNAL $rest` instead of typing in passwords with `ldapadd -x -D cn=admin,dc=example,dc=com -W $rest`
  - now we can replace all ldapadd/modify/... references in this guide binding to `cn=admin,dc=example,dc=com` with this root command 😸 
  - we can remove the "The command will ask for the admin password" reference.
  - we can link to the "SASL identity" heading in the acl section for further explanation of this mechanism
- "Change the configuration"
  - mention that usually the ldap configuration is within the database, so it can be replicated. the config can also be done with files in `/etc/ldap/slapd.d`
- "Add an Index":
  - "You can confirm the change with a search:" -> the next `ldapsearch` command querying for `olcDbIndex` needs sudo
- "Add a schema":
  - since currently there's no mention or reference, it could benefit from a short explanation what a schema is - "describes the structure and data types of available object classes in the ldap tree" 
- "Logging":
  - we prefer systemd journal logging now.
  - mention viewing the logs with `sudo journalctl -eu slapd.service`
  - rate limiting with rsyslog should be removed, we could instead mention [how this would be changed in journald](https://www.freedesktop.org/software/systemd/man/latest/journald.conf.html#RateLimitIntervalSec=) or just leave it out since this is unlikely to be a problem that should be resolved by changing the limits in practice.

### access control
- meta description is cut off and generally not well fitting
- commands should be in separate code blocks
- there's no mention of how a acl is set and created
- we should set an example how to do **passwordless authentication**, otherwise we'd have to type a password for many many of the tutorial commands. this is done by creating `peercred.ldif`:
```ldif
# allow root user (via peercred) to manage the primary database (mdb)
dn: olcDatabase={1}mdb,cn=config
changetype: modify
add: olcAccess
# inserted as first acl rule
olcAccess: {0}to * by dn.exact="gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth" manage by * break
```
- then activating this new access config via `sudo ldapmodify -H ldapi:/// -Y EXTERNAL < peercred.ldif`
- fix the "access control topic" reference link to point to release 2.6, i.e. https://openldap.org/doc/admin26/guide.html#Access%20Control

### replication
- references should point to the 2.6 release, currently it's 2.4.
- the meta description isn't wrong, but could be more precise to just say "set up data synchronization for having a cluster of ldap servers for high availability"
- code blocks and output should again be separate (and no included $ when copying the code)
- should also use the newly setup passwordless peercred acl instead of entering admin passwords for all changes.

### LDAP and TLS
- meta description too verbose and cuts off
- code blocks should again be separate for command and output, so we can copy it and don't include `$ ` when copying
- although a manual certificate setup is nice for understanding, it should primarily explain how to use let's encrypt for certificates. since many other articles will have the same base, we should have a general let's encrypt setup guide, where we refer to for most of the setup, and just keep the openldap specific bits in here. also, the ca setup is duplicated in many other setup guides, so we should also move the manual ca setup to a central shared document.
- the note "a call to dpkg-reconfigure ca-certificates is necessary." is unclear if one needs to run this command. after testing i can confirm you don't need to run it. therefore the description should say "in case you want to select ca certificates selectively, you can run ..."
- this duplicates content in https://documentation.ubuntu.com/server/how-to/security/install-a-root-ca-certificate-in-the-trust-store/ or https://documentation.ubuntu.com/server/explanation/security/certificates/ which should remain the primary sources.
- "Certificate for an OpenLDAP replica"
  - this should not duplicate the certificate generation again, and not be part of the general tls page
  - it should rather be part of the replication article, and refer to the common part in the "ldap and tls" article from there (this also allows us to prefer using let's encrypt instead, while keeping the manual setup for reference)

### Backup and restore OpenLDAP

- should use systemd timers instead of cron jobs
- the backup script can be improved to this:
```bash
#!/bin/bash
set -euo pipefail

BACKUP_PATH=/export/backup

# backup server config
nice slapcat -b cn=config > ${BACKUP_PATH}/config.ldif
# backup directory tree
nice slapcat -b dc=example,dc=com > ${BACKUP_PATH}/example.com.ldif
chmod 600 ${BACKUP_PATH}/*.ldif

# if needed, use borgbackup or similar to store the backups somewhere safe
```
- the restore script shouldn't hardcode the full path of `SLAPADD=/usr/sbin/slapadd`, instead just call `slapadd` directly.

TheJJ:
### Description

Document a simple but robust way to set up a LDAP client for an existing LDAP server.

Additionally, i fixed the stuff found in #585 

---

### Related Issue

fixes #585 

### Checklist

- [ ] I have tested my changes, and they work as expected (`Could not import extension sphinx_llm.txt (exception: No module named 'sphinx_llm')`).

### DIFF
--- a/docs/how-to/openldap/access-control.md
+++ b/docs/how-to/openldap/access-control.md
@@ -1,12 +1,11 @@
 ---
 myst:
   html_meta:
-    description: Configure OpenLDAP access control lists (ACLs) to manage read, write, and authentication permissions for LDAP resources.
+    description: Configure OpenLDAP access control lists (ACLs) to manage read, write, and authentication permissions.
 ---
 
 (ldap-access-control)=
 # Set up LDAP access control
-
 
 The management of what type of access (read, write, etc) users should be granted for resources is known as **access control**. The configuration directives involved are called **access control lists** or {term}`ACL`s.
 
@@ -16,20 +15,32 @@ To get the effective ACL for an LDAP query we need to look at the ACL entries of
 
 ## Getting the ACLs
 
-The following commands will give, respectively, the ACLs of the `mdb` database (`dc=example,dc=com`) and those of the frontend database:
+The following commands give the ACLs of the `mdb` database (`dc=example,dc=com`) and those of the frontend database respectively.
+
+For the main database:
 
 ```bash
-$ sudo ldapsearch -Q -LLL -Y EXTERNAL -H ldapi:/// -b \
-cn=config '(olcDatabase={1}mdb)' olcAccess
-    
+sudo ldapsearch -Q -LLL -Y EXTERNAL -H ldapi:/// -b cn=config '(olcDatabase={1}mdb)' olcAccess
+```
+
+Output:
+
+```text
 dn: olcDatabase={1}mdb,cn=config
 olcAccess: {0}to attrs=userPassword by self write by anonymous auth by * none
 olcAccess: {1}to attrs=shadowLastChange by self write by * read
 olcAccess: {2}to * by * read
+```
 
-$ sudo ldapsearch -Q -LLL -Y EXTERNAL -H ldapi:/// -b \
-cn=config '(olcDatabase={-1}frontend)' olcAccess
-    
+For the frontend database:
+
+```bash
+sudo ldapsearch -Q -LLL -Y EXTERNAL -H ldapi:/// -b cn=config '(olcDatabase={-1}frontend)' olcAccess
+```
+
+Output:
+
+```text
 dn: olcDatabase={-1}frontend,cn=config
 olcAccess: {0}to * by dn.exact=gidNumber=0+uidNumber=0,cn=peercred,cn=external
  ,cn=auth manage by * break
@@ -57,7 +68,7 @@ to attrs=userPassword
     by self write
     by anonymous auth
     by * none
-    
+
 to attrs=shadowLastChange
     by self write
     by * read
@@ -87,15 +98,18 @@ If this is unwanted then you need to change the ACL. To force authentication dur
 There is no administrative account ("Root DN") created for the `slapd-config` database. There is, however, a SASL identity that is granted full access to it. It represents the localhost's superuser (`root`/`sudo`). Here it is:
 
 ```text
-dn.exact=gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth 
+dn.exact=gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth
 ```
 
-The following command will display the ACLs of the `slapd-config` database:
+The following command displays the ACLs of the `slapd-config` database:
 
 ```bash
-$ sudo ldapsearch -Q -LLL -Y EXTERNAL -H ldapi:/// -b \
-cn=config '(olcDatabase={0}config)' olcAccess
-    
+sudo ldapsearch -Q -LLL -Y EXTERNAL -H ldapi:/// -b cn=config '(olcDatabase={0}config)' olcAccess
+```
+
+Output:
+
+```text
 dn: olcDatabase={0}config,cn=config
 olcAccess: {0}to * by dn.exact=gidNumber=0+uidNumber=0,cn=peercred,
               cn=external,cn=auth manage by * break
@@ -107,11 +121,40 @@ Since this is a SASL identity we need to use a SASL **mechanism** when invoking 
 
 - The EXTERNAL mechanism works via **Inter-process Communication** (IPC, UNIX domain sockets). This means you must use the `ldapi` URI format.
 
-A succinct way to get all the ACLs is like this:
+(ldap-peercred-setup)=
+## Grant root passwordless access to the data tree
+
+By default, the local root user can manage `cn=config` via SASL EXTERNAL, but not the data tree (`dc=example,dc=com`). To allow root to also manage the data tree without entering a password, add an ACL rule.
+
+Create a file called `peercred.ldif`:
+
+```text
+# Allow root user (via peercred) to manage the primary database (mdb)
+dn: olcDatabase={1}mdb,cn=config
+changetype: modify
+add: olcAccess
+# Inserted as the first ACL rule
+olcAccess: {0}to * by dn.exact="gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth" manage by * break
+```
+
+Apply the change:
 
 ```bash
-$ sudo ldapsearch -Q -LLL -Y EXTERNAL -H ldapi:/// -b \
-cn=config '(olcAccess=*)' olcAccess olcSuffix
+sudo ldapmodify -Q -Y EXTERNAL -H ldapi:/// -f peercred.ldif
+```
+
+Now you can run LDAP commands to modify the data tree, as root, without a LDAP password (when you run this on the OpenLDAP server machine):
+
+```bash
+sudo ldapadd -Q -Y EXTERNAL -H ldapi:/// -f add_content.ldif
+```
+
+This simplifies administration and is used throughout the other guides in this series.
+
+A succinct way to get all the ACLs is:
+
+```bash
+sudo ldapsearch -Q -LLL -Y EXTERNAL -H ldapi:/// -b cn=config '(olcAccess=*)' olcAccess olcSuffix
 ```
 
 ## Next steps
@@ -121,4 +164,4 @@ See how to {ref}`set up LDAP users and groups <ldap-users-and-groups>`.
 ## Further reading
 
 - See the manual page for {manpage}`slapd.access(5)`
-- The [access control topic](https://openldap.org/doc/admin25/guide.html#Access%20Control) in the OpenLDAP administrator's guide.
+- The [access control topic](https://openldap.org/doc/admin26/guide.html#Access%20Control) in the OpenLDAP administrator's guide.
--- a/docs/how-to/openldap/backup-and-restore.md
+++ b/docs/how-to/openldap/backup-and-restore.md
@@ -1,65 +1,115 @@
 ---
 myst:
   html_meta:
-    description: Back up and restore OpenLDAP databases including configuration backend and DIT using slapcat and shell scripts.
+    description: Back up and restore OpenLDAP configuration and data using slapcat and systemd timers.
 ---
 
 (ldap-backup-and-restore)=
 # Backup and restore OpenLDAP
 
-Now we have LDAP running just the way we want, it is time to ensure we can save all of our work and restore it as needed.
+Once you have LDAP running the way you want, it is time to ensure you can save all your work and restore it as needed.
 
-What we need is a way to back up the directory database(s) -- specifically the configuration backend (`cn=config`) and the {term}`DIT` (`dc=example,dc=com`). If we are going to backup those databases into, say, `/export/backup`, we could use `slapcat` as shown in the following script, called `/usr/local/bin/ldapbackup`:
+What we need is a way to back up the directory database(s) -- specifically the configuration backend (`cn=config`) and the {term}`DIT` (`dc=example,dc=com`).
+
+## Backup script
+
+Create `/usr/local/bin/ldapbackup` with the following content:
 
 ```bash
 #!/bin/bash
-
-set -e
+set -euo pipefail
 
 BACKUP_PATH=/export/backup
-SLAPCAT=/usr/sbin/slapcat
+CONFIG_BACKUP=f${BACKUP_PATH}/config.ldif"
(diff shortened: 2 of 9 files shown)


docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,264 @@
...
+
+The installer updates `/etc/pam.d/common-*` automatically. To confirm that LDAP authentication and automatic home directory creation are enabled, run:
+
+```bash
+sudo pam-auth-update
+```
+
+Select **LDAP Authentication** and **Create home directory on login** in the interactive menu.
+
+## Restart the services
```
s-makin: We don't need a header for this - restarting the service is part of the configuration step

docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,264 @@
...
+
+List all groups:
+
+```bash
+getent group
+```
+
+If the output includes users and groups from the LDAP directory, NSS integration is working correctly.
+
+## Debug nslcd
```
s-makin: This seems out of place - usually debugging/troubleshooting goes at the bottom of the page. But the title and the section seem to be mismatched, because inspecting live queries using a debug mode doesn't seem (to me) to be the same thing as debugging the service. I'll let you decide whether to move the section or tweak the heading!

docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,264 @@
...
+```{note}
+Local users not in any listed LDAP group can be permitted to log in by adding them to a local `login` group. Create the group and add users as needed:
+
+    sudo groupadd -r login
+    sudo usermod -aG login <username>
+```
+
+## Grant sudo access to an LDAP group
+
+To allow members of an LDAP group `sudo` access without a password, add an entry to `/etc/sudoers` using {manpage}`visudo(8)`:
```
s-makin: aren't we using sudors now?

s-makin (approved):
really nice changes overall, only a couple queries from me. Pre-approving.

docs/how-to/openldap/ldap-client.md
TheJJ: I made it part of the "test the configuration" section.

docs/how-to/openldap/ldap-client.md
TheJJ: we are, but this instructs how depending on ldap group membership you're allowed to use sudo then. i've improved the wording!

TheJJ:
Updated :)

s-makin (approved):
All lgtm, thanks for the changes! I don't know if you're waiting on review from anyone else so I won't merge, but approving.

docs/how-to/openldap/access-control.md
```diff
@@ -107,11 +121,40 @@ Since this is a SASL identity we need to use a SASL **mechanism** when invoking
...
+olcAccess: {0}to * by dn.exact="gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth" manage by * break
+```
+
+Apply the change:
+
+```bash
+sudo ldapmodify -Q -Y EXTERNAL -H ldapi:/// -f peercred.ldif
+```
+
+Now you can run LDAP commands as root without passwords:
```
panlinux: "Now you can run LDAP commands to modify the data tree, as root, without a password".

panlinux: Also "as long as they are run on the same server where OpenLDAP is installed, of course" (maybe written in a different manner). Optional, as it might be obvious.

docs/how-to/openldap/backup-and-restore.md
```diff
@@ -1,65 +1,110 @@
...
+
+Make it executable:
+
+```bash
+sudo chmod +x /usr/local/bin/ldapbackup
 ```
 
 ```{note}
-These files are uncompressed text files containing everything in your directory including the tree layout, usernames, and every password. So, you might want to consider making `/export/backup` an encrypted partition and even having the script encrypt those files as it creates them. Ideally you should do both, but that depends on your security requirements.
+These files are uncompressed text files containing everything in your directory including the tree layout, usernames, and every password. Consider making `/export/backup` an encrypted partition and even having the script encrypt files as it creates them.
```
panlinux, line 45: At the very least, the proposed backup script should take care of adjusting the permissions of the backups 0600. I would even suggest to create the file empty first, chmod, and then run the backup, because slapcat could take a few seconds to run, depending on the size of the tree.

TheJJ: `chmod 600 ${BACKUP_PATH}/*.ldif` is in there? good idea with the empty creation, added.

TheJJ:
Thx, good ideas, added.

docs/how-to/openldap/backup-and-restore.md
panlinux: Oops, I missed it then.

docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,274 @@
...
+  html_meta:
+    description: Configure an Ubuntu system as an LDAP client using nslcd and NSS to make directory users and groups available for local authentication.
+---
+
+(ldap-client)=
+# Set up an LDAP client
+
+This guide explains how to configure an Ubuntu machine as a [Lightweight Directory Access Protocol (LDAP)](https://en.wikipedia.org/wiki/Lightweight_Directory_Access_Protocol) client, so that users and groups stored in an LDAP directory become available on the system for authentication. This enables users to log in via SSH or other services backed by [Pluggable Authentication Modules (PAM)](https://en.wikipedia.org/wiki/Pluggable_Authentication_Module) using their LDAP credentials.
+
+The approach used here is `nslcd` with `libnss-ldapd` and `libpam-ldapd`. The Name Service Switch (NSS) and PAM modules communicate with the `nslcd` daemon, which maintains a single shared LDAP connection rather than each process opening its own. This keeps the setup lightweight and easy to debug.
```
panlinux: Maybe link NSS to https://ubuntu.com/server/docs/reference/glossary/#term-NSS. It's not filled in yet, though.

docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,274 @@
...
+```bash
+sudo apt install nslcd libpam-ldapd
+```
+
+The installer prompts for your LDAP server URI and base DN. For example:
+
+- **LDAP server URI:** `ldaps://ldap.example.com`
+- **Base DN:** `dc=example,dc=com`
+
+Installing `libnss-ldapd` (pulled in as a dependency) also updates `/etc/nsswitch.conf` to include `ldap` as a source for `passwd`, `group`, and `shadow` lookups.
```
panlinux: `libpam-ldapd` is also a Recommends in `nslcd`, any reason why you included it in the apt install command, and not libnss-ldapd?

docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,274 @@
...
+sudo systemctl restart nslcd
+```
+
+## Test the configuration
+
+Verify that a known LDAP user is visible to the system:
+
+```bash
+id <username>
+```
```
panlinux, line 81: I think the prerequisites should also mention how the DIT in openldap should be populated, i.e., what's the schema for users and groups. Definitely not worth it explaining how that is done here, but a pointer is needed. I believe we are talking about the RFC2307 schema, right? Perhaps link to https://ubuntu.com/server/docs/how-to/openldap/users-and-groups/ (I haven't checked if `ldapscripts` still works, though...). We mention such requirements similarly (RFC2307) in https://ubuntu.com/server/docs/how-to/sssd/with-ldap/ and https://ubuntu.com/server/docs/how-to/sssd/with-ldap-and-kerberos/

docs/how-to/openldap/ldap-client.md
TheJJ: reworded and added to main `apt install` command to make sure users install it.

docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,274 @@
...
+sudo systemctl stop nslcd.service
+sudo nslcd -n -d
+```
+
+Press {kbd}`Ctrl+C` to stop this debug session.
+
+You can iterate with configuration or setup adjustments until the log output and `getent` results are satisfying.
+
+
+Finally, restart the service via {ref}`systemd`:
```
panlinux: It could be confusing to say to restart it with "systemd", and then you use the "systemctl" command ;)

docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,274 @@
...
+
+```bash
+sudo systemctl start nslcd.service
+```
+
+
+## Configure home directories
+
+By default, `nslcd` uses the `homeDirectory` attribute from the LDAP directory. You can remap this — or any other `passwd` field — in `/etc/nslcd.conf` using `map` directives. For example, to construct home directory paths locally regardless of what the LDAP directory provides:
```
panlinux: I think /etc/passwd has a section 5 manpage, explaining the fields. If it does (I didn't check just yet), it would be a better reference than the passwd (1) command.

docs/how-to/openldap/ldap-client.md
TheJJ: done by crosslinking

docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,274 @@
...
+        required    pam_access.so nodefgroup
+Account-Type: Primary
+Account:
+        required    pam_access.so nodefgroup
+Session-Type: Additional
+Session:
+        required    pam_access.so nodefgroup
+```
+
+Apply the configuration to add these modules to `/etc/pam.d/common-*`:
```
panlinux, line 161: I was surprised we didn't have this enabled by default somehow. I just checked, and we have commented entries for pam_access.so in several /etc/pam.d/* files. The above looks cleaner than uncommenting entries by hand, for sure.

docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,274 @@
...
+```bash
+sudo pam-auth-update --enable ldap-access --enable ldap --enable mkhomedir
+```
+
+```{note}
+Local users not in any listed LDAP group can be permitted to log in by adding them to a local `login` group. Create the group and add users as needed:
+
+    sudo groupadd -r login
+    sudo usermod -aG login <username>
+```
```
panlinux, line 174: Is this `login` group a magical group that automatically grants that kind of access? Perhaps in the default /etc/security/access.conf?

docs/how-to/openldap/ldap-client.md
```diff
@@ -0,0 +1,274 @@
...
+```text
+%your-admin-group   ALL=(ALL) ALL
+```
+
+To allow `sudo` usage without entering a password, add this entry instead using {manpage}`visudo(8)`:
+```text
+%your-admin-group   ALL=(ALL) NOPASSWD: ALL
+```
+
+Replace `your-admin-group` with the relevant LDAP group name.
```
panlinux, line 191: Perhaps add a note that this is no different from granting sudo access to a local group.

docs/how-to/openldap/ldap-client.md
TheJJ: this is the group we specify in `access.conf` to have access. could be any name.

panlinux (approved):
+1

PR #713, opened 2026-05-27
Tutorial pt 2 - CLI in depth

s-makin:
### Description

Part 2 of aligning our content with our external training partners. This tutorial covers the more advanced basic terminal instructions.

Also added the commands covered to the CLI reference page, and added links to this tutorial from part 1.

---

### Checklist

---

---


docs/tutorial/cli-in-depth.md
```diff
@@ -0,0 +1,778 @@
...
+-rwxr-xr-x 1 root root      174328 Apr 30 13:32 kmod
+lrwxrwxrwx 1 root root           4 Apr 30 13:32 lsmod -> kmod
+```
+
+Each command in the pipeline only ever sees the output of the command immediately before it -- the shell connects them all behind the scenes.
+
+
+## Chaining commands
+
+Pipes pass *data* from one command to another. Command chaining is different: it controls the *order* in which commands run, and whether the next command runs at all.
```
grayson-wolf:
```suggestion
Pipes pass *data* from one command to another and run all of their commands in parallel. This is to allow commands begin working on long streams of data while their source is still producing more. While often useful, you might not want that, in which case you would use **command chaining**.  Chaining is different: it controls the *order* in which commands run, and whether the next command runs at all.
```

Since we're talking about command order here, I think we should explicitly mention that pipe limbs run in parallel. 

I could be convinced that this warrants an even greater mention than this (since this has definitely caused me headache at least once when writing shell scripts), but for now I think mentioning it here is good.

docs/tutorial/cli-in-depth.md
```diff
@@ -0,0 +1,778 @@
...
+:host: tutorial
+:dir: ~
+ls; pwd; whoami
+
+errors.txt  file-list.txt  greeting.txt  output.txt
+/home/ubuntu
+ubuntu
+```
+
+Often we want the second command to run *only* if the first one succeeded. We use `&&` for this:
```
grayson-wolf, line 334:
```suggestion
On Linux, the way processes in general declare whether they have "succeeded" or not is with an **exit code.** An exit code (also interchangeably called "exit status") is a number from zero to 255 that a process gives to its parent upon completion. "Success" is defined by exiting with code `0`, with *all* other values indicating (different types of) failure. These might be specific to the program, but in general your generic failure status will be exit code `1`. You can use `exit n` to immediately exit with code `n`.

Often we want the second command to run *only* if the first one succeeded. We use `&&` for this:
```

Since we're talking about "success" now, we should clarify what that means. There's more info later, but I think a few sentences here is fine. 

(tangent: git is awfully ambitious with where it's placing the diff on this particular comment :) )

docs/tutorial/cli-in-depth.md
```diff
@@ -0,0 +1,778 @@
...
+[...]
+Host is up
+```
+
+A few other operators are useful to know. Try each of these out, both with and without the operators, to see how they change the behaviour for yourself:
+
+Background execution; `&`
+: Runs a command in the background. The shell prompt returns immediately, and the command continues running, e.g.:
+: ```text
+  nano file-list.txt &
```
grayson-wolf: `nano` is probably not the right choice of example for this, since a text editor is probably something you pretty much always want actually open, and backgrounding it doesn't really make sense. 

The example that immediately comes to mind is something like using `wget` to fetch a large file (say, an Ubuntu Server ISO 😉):
```
wget https://cdimage.ubuntu.com/ubuntu-server/resolute/daily-live/current/resolute-live-server-amd64.iso
```
but also I don't think having the example command fetch a >1G file is a good idea.  

`sleep` doesn't actually *do* anything, but it's probably still the most useful example. Maybe have them do `sleep 10`, observe that their shell prompt is gone for 10 seconds, then `sleep 10 &` and viola (🎻 )! shell remains accessible! 

In any case, it's probably worth mentioning `&` outputs the job number and PID of the backgrounded process:
```
grayson.wolf@canonical.com@penemue:~$ sleep 10 &
[1] 235741
```
if only so people know What The Weird Numbers Mean or if they need to `kill` it later.

docs/tutorial/cli-in-depth.md
```diff
@@ -0,0 +1,778 @@
...
+: ```text
+  { echo "System Report"; date; uptime; } > report.txt
+  ```
+
+Subshell execution; `()`
+: Run commands in a subshell, where the commands are forgotten the moment they finish running. This is useful for isolating side effects like a `cd` -- the directory change does not affect the current shell.
+: ```text
+  (export NAME="Alice"; echo "Hello $NAME")
+  echo "Name is: $NAME"
+  ```
```
grayson-wolf, line 399: I think this makes more sense to be moved into the section on env vars (maybe just calling it "The Environment" in general), because at least in my case the most common usage of this kind of thing is capturing output like `GITHUB_TOKEN="$(gh auth token)"`, and since the important part about `()` is that it doesn't change the environment (and since your example uses env vars!)

docs/tutorial/cli-in-depth.md
```diff
@@ -0,0 +1,778 @@
...
+
+Line continuation; `\`
+: Continue a long command on the next line, which helps a lot with command readability.
+
+The `()` subshell is particularly handy when we want to work in a different directory without losing our current location:
```
grayson-wolf: should this have a command with it? 

in any case it seems separated from the rest of the info by the line continuation bit, so this should move.

docs/tutorial/cli-in-depth.md
```diff
@@ -0,0 +1,778 @@
...
+Hello from the terminal!
+This text is going into a file.
+```
+
+
+## Pipes
+
+Pipes are one of the most powerful ideas in the Unix philosophy: the output of one command becomes the input of the next. We connect them with the `|` operator (the "pipe" character).
+
+For example, `ls -la /usr/bin` produces many lines of output -- more than can fit on screen, and you can only see the end of the output (you can't scroll back to the top). To make it more convenient, we can pipe the output to `less` so we can page through it with the {kbd}`space` key:
```
grayson-wolf:
```suggestion
For example, `ls -la /usr/bin` produces many lines of output -- more than can fit on screen, and you can only see the end of the output (scrolling back up is very inconvenient, if you can even do it at all (some terminals can't!)). To make it more convenient, we can pipe the output to `less` so we can page through it with the {kbd}`space` key:
```

nit: i think most modern terminal emulators actually can scroll back up, so I reworded this

docs/tutorial/cli-in-depth.md
```diff
@@ -0,0 +1,778 @@
...
+:user: ubuntu
+:host: tutorial
+:dir: ~
+DAY="Monday"
+echo "Today is ${DAY}!"
+
+Today is Monday!
+```
+
+Without the braces, the shell would try to expand `$DAY!` as a single variable name and find nothing.
```
grayson-wolf, line 664:
```suggestion
Without the braces, the shell would try to expand `$DAY!` as a single variable name and find nothing.

It's also important to note that this variable expansion applies *everywhere* in the line, even with the command name itself!:
\`\`\`{terminal}
:copy:
:user: ubuntu
:host: terminal
:dir: ~
CMD="echo"
$CMD "It's dynamic!"
It's dynamic!
\`\`\`
This can be useful in, say, a script that runs one of two very similar commands with the same argument, so you don't have to re-type everything twice in your conditional.
```

Added a brief caveat about using variables as command names (you might need to remove the anti-github-markdown backslashes).

grayson-wolf (changes requested):
Looks mostly good, just some nits and handful of rearranging proposals and questions about the examples.

docs/tutorial/welcome-to-the-terminal.md
```diff
@@ -17,7 +17,6 @@ In this tutorial, we'll use a virtual machine (VM) with Ubuntu Server to poke ar
 
   The Multipass virtual machine needs at least **5 GB of disk space**, and **1 GB of memory**.
 
-
 (tutorial-install-multipass)=
 ## Install Multipass
```
cpaelzer: We get into the redudancy weirdness.
- Welcome to the terminal has this short section
  - "explaining how" - but short and mostly by referring to upstream docs
- CLI in depth
  - "explaining why" - knowledge and hardware section if we explain it here why not in the other?
    - outlining that one will need a virtual machine but saying hard "use mutlipass". Generally this should be "you can use a Ubuntu VM whichever way you like, we will in these examples use Multipass to create"
    - Only then "explaining how" - but short and mostly by referring to upstream docs
- Managing your software
  - "explaining why": has a slightly different knowledge & hardware section - why++?
  - "explaining how" - in long details with steps and then how to use it to create the VM. having the details on the last tutorial is unexpected :-P


IMHO all these should be written ONCE with same style and content - and then either
a) only be a link to a dedicate page about it or
b) included in each of the tutorial pages for rendering to be the very same in each.

cpaelzer (changes requested):
The more tutorials we get (good) the less I like the multipath intro

docs/tutorial/cli-in-depth.md
s-makin: good catch! adopted

docs/tutorial/cli-in-depth.md
s-makin: I actually didn't know they ran in parallel, so once again TIL :D 

I've adopted your wording, with some tweaks. I probably won't expand on it in this tutorial - if we ever write one about scripting, that would be the best place to expand on it imo. Maybe I'll do one for the new wiki?

docs/tutorial/cli-in-depth.md
s-makin: Didn't know this either - it seems like an interesting aside so I've called it out in an admonition block rather than just body text.

docs/tutorial/cli-in-depth.md
s-makin: I use `gedit <file> &` all the time 😅 this is what happens when you get taught to Linux by *not* computer science people, you get bad habits

Sleep is a nice example though, thanks! Adding both that and the PID note.

docs/tutorial/cli-in-depth.md
s-makin: Ah, no this is an artifact of my editing style. Anytime I'm not happy with the wording I redraft on multiple lines - sometimes I forget to delete the discarded ones lol. deleted it now.

docs/tutorial/cli-in-depth.md
s-makin: Yeah this is a bit of a chicken-and-egg situation. I decided that since several operators are used in the env vars section (not just the subshell execution), they should be specifically introduced first as a logical/conceptual group so that a self-reinforcing loop is made when it's actually used in examples for useful things later, without necessarily coupling them together in the mind of the reader (implying more restricted usage than reality allows).

docs/tutorial/welcome-to-the-terminal.md
s-makin: I agree, it was becoming unbearable. Since any of the tutorials could be an entry point, I've extracted the multipass step into a new txt file that is now included in each tutorial. Since it's potentially repetitious for anyone doing multiple tutorials, I also added links to skip to the content.

s-makin:
I'm going to postpone merging this until next week. Given that I have now updated the multipass instructions to use resolute everywhere, the code blocks in the managing software tutorial need to be updated to follow, which will take me some time to test.

docs/reference/cli-cheatsheet.md
```diff
@@ -82,6 +84,64 @@ For full details on any command, run `man <command>` in your terminal.
...
+| `(cmd1; cmd2)` | Run commands in a subshell (side effects don't affect the current shell) |
+| `cmd1 \` | Continue a long command on the next line |
+
+
+## Variables and environment
+
+| Command | Description | Online manual page |
+|---------|-------------|--------------------|
+| `env` | List all environment variables |{manpage}`env(1)`|
+| `echo $VAR` | Print the value of variable `VAR` |{manpage}`echo(1)`|
```
guilherme-puida: nitpicky, but `echo` is a shell builtin in bash, dash and zsh.

```
% for s in bash dash zsh; do $s -c 'type echo'; done
echo is a shell builtin
echo is a shell builtin
echo is a shell builtin
```

This is relevant because `/usr/bin/echo` (which `echo(1)` documents) behaves differently when compared to the `echo` builtin:

```
% echo --help
--help

% /usr/bin/echo --help
Usage: /usr/bin/echo [SHORT-OPTION]... [STRING]...
  or:  /usr/bin/echo LONG-OPTION
Echo the STRING(s) to standard output.

  -n             do not output the trailing newline
  -e             enable interpretation of backslash escapes
  -E             disable interpretation of backslash escapes (default)
      --help        display this help and exit
      --version     output version information and exit

If -e is in effect, the following sequences are recognized:

  \\      backslash
  \a      alert (BEL)
  \b      backspace
  \c      produce no further output
  \e      escape
  \f      form feed
  \n      new line
  \r      carriage return
  \t      horizontal tab
  \v      vertical tab
  \0NNN   byte with octal value NNN (1 to 3 digits)
  \xHH    byte with hexadecimal value HH (1 to 2 digits)

NOTE: your shell may have its own version of echo, which usually supersedes
the version described here.  Please refer to your shell's documentation
for details about the options it supports.

NOTE: printf(1) is a preferred alternative,
which does not have issues outputting option-like strings.

GNU coreutils online help: <https://www.gnu.org/software/coreutils/>
Full documentation <https://www.gnu.org/software/coreutils/echo>
or available locally via: info '(coreutils) echo invocation'
```

docs/tutorial/cli-in-depth.md
s-makin: Added!

docs/tutorial/welcome-to-the-terminal.md
cpaelzer: I see they are now defined once and equally displayed - +1 on mine.

docs/reference/cli-cheatsheet.md
cpaelzer: maybe point to https://manpages.ubuntu.com/manpages/resolute/man1/bash.1.html#shell-builtin-commands instead?

cpaelzer: or even https://manpages.ubuntu.com/manpages/resolute/man1/bash.1.html#echo~2

cpaelzer (approved):
Thanks for the fixups, the remaining two topics can be sorted out in follow up PRs if you want

docs/reference/cli-cheatsheet.md
s-makin: Another TIL, thanks @guilherme-puida - I only ever used bash, and never even looked at the manpage lol. I'll change that in this PR, since I'm touching the page anyway.

grayson-wolf (approved):
Changes look good to me!

s-makin:
Thanks all for your reviews! Much obliged. I'm going to merge this now, and we can continue to iterate in followup PRs.

Issue #129, opened 2025-01-31
Update: building up the new glossary


s-makin:
## The problem

Now that [we have a glossary](https://github.com/canonical/ubuntu-server-documentation/pull/124), we need to begin defining the terms. 

## What is needed

Since the glossary was based on the spelling exceptions list I have made a first pass to clear out the spelling exceptions, duplicates, and plural items. However, there are still plenty of terms that don't belong there, and all terms need to be defined. Some terms/acryonyms may have more than one definition. 

- Each *valid* term in the glossary needs a clear definition.
- Invalid terms need to be removed.
- Once a term has a definition, it needs to be referencedlinked in any pages that use that term (at the first place on each page that the term appears).

**What's a valid term?**

- Acronyms
- Product or package names (definition of the spelling and capitalization should come from the official product documentation)
- Technical terms/jargon

**What's an invalid term?**

- Code elements/terms that only appear in code blocks without a corresponding textual definition
- A plural of a term that's already been defined
- A duplicate of a term with alternative spelling
- Terms that *only* appear in the glossary page

## Suggestions

I don't recommend trying to tackle the entire list in one PR, it's very long and would be a lot for one person to do alone! Pick a subset of terms you want to define (I suggest no more than 5-10 in one go).
 
- For each term, search in the [rendered documentation](https://documentation.ubuntu.com/server/) to see which page(s) it appears in - you'll want to make a note of those for later.

  - If it's an invalid term, it can be deleted from the glossary list. 
  - If it only has a single definition/meaning, include the definition in the glossary - it's fine to pick this up from the documentation. Feel free to link to other supporting material (such as manual pages or official package documentation) as you see fit.
  - It it has more than one meaning, make sure to include all possible definitions.

- It would be great to also refer to the section/topic the term relates to (e.g. ACLs are related to security, but also pops up in the OpenLDAP and Kerberos content). This will help in the future, as we may want to provide section-specific glossaries, or organise the glossary into sections.
- Finally, after you have defined the term, include a reference to the glossary entry on *each page* where the term appears, using the MyST syntax:
  ```
  {term}`glossary term to link to`
  ```
  The term inside the backticks needs to match the glossary term *exactly*. But, if you want to point to a term with a different link text, you can use the following syntax instead:
  ```
  {term}`Acess Control Lists <ACL>`
  ```
  Where in this example, "Acess Control Lists" is the link text shown to the user, and ACL is the term as it appears in the glossary.
[...]

PR #717, opened 2026-05-29
Add glossary definitions for IP, IP address, iSCSI, localhost and LTS

drkrillo:
### Description

Adds definitions for IP, IP address, iSCSI, localhost and LTS to the glossary.

---

### Related Issue

Part of #129

---

### Commit Message for Squash Merge

Add glossary definitions for IP, IP address, iSCSI, localhost and LTS

---

### Checklist

---

I'll do a follow up PR to link the glossary terms to the first occurrence in the other doc pages when named.

### DIFF
--- a/docs/reference/glossary.md
+++ b/docs/reference/glossary.md
@@ -1304,10 +1304,25 @@ Internet of Things
 
 IP
 Internet Protocol
-    *Work in Progress*
+    IP is a connectionless network protocol responsible for routing packets across
+    networks. It uses the IP {term}`datagram <Datagram>` as its basic unit of
+    network information.
+
+    See also:
+    * [Networking Key Concepts](https://ubuntu.com/server/docs/explanation/networking/networking-key-concepts/)
+
+    Related topic(s):
+    * Networking
 
 IP address
-    *Work in Progress*
+    An IP address is a unique numerical identifier for a device on a network. Common
+    versions include IPv4 (32-bit, i.e. 192.168.1.1) and IPv6 (128-bit).
+
+    See also:
+    * [Networking Key Concepts](https://ubuntu.com/server/docs/explanation/networking/networking-key-concepts/)
+
+    Related topic(s):
+    * Networking
 
 IPC
 Inter-Process Communication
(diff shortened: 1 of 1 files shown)


docs/reference/glossary.md
```diff
@@ -1304,10 +1304,25 @@ Internet of Things
 
 IP
 Internet Protocol
-    *Work in Progress*
+    IP is a connectionless network protocol responsible for routing packets across
+    networks. It uses the IP {term}`datagram <Datagram>` as its basic unit of
+    network information.
+
+    See also:
+    * [Networking Key Concepts](https://ubuntu.com/server/docs/explanation/networking/networking-key-concepts/)
```
s-makin: For internal pages within the Server docs, we use a special MyST syntax that lets us point to an anchor on the target page, rather than hardcoding the link to the page itself. This means that if a page gets moved, it doesn't break the link and we don't have to update it. I try to keep the anchor names the same as the page/file names to make identifying the anchors easier. In this case, instead of using the usual markdown syntax of []() for the link, we'd use {ref}`networking-key-concepts`. Then when the docs are built, the builder cross-references the anchor to pull in the page title which is what gets shown on the page, so we don't even need to specify the title - very handy!

I'll add a suggestion here for this one so you can see clearly what the syntax should look like. Please can you then update all the others to follow the same pattern? :)

s-makin:
```suggestion
    * {ref}`networking-key-concepts`
```

docs/reference/glossary.md
```diff
@@ -1493,7 +1523,13 @@ Logic Systems Incorporated
...
 LTS
 Long-Term Support
-    *Work in Progress*
+    LTS is a release policy that provides a stable version of a product for an
+    extended period of time. For example, Ubuntu LTS releases receive 5 years of
+    standard support, including security patches, bug fixes and hardware
+    compatibility updates, prioritizing stability over newer package versions.
+
+    See also:
+    * {term}`Expanded Security Maintenance <ESM>`
```
s-makin: I think this would be a good place to link to the Ubuntu Project releases pages, which explains in more detail about LTS, how releases work etc. Can we replace the term link with this URL instead? https://documentation.ubuntu.com/project/release-team/ubuntu-releases/

s-makin (commented):
Thanks for your patience @drkrillo, I've finally been able to take a look at your PR - this is a really nice contribution, thanks! The definitions look great, I really like how concise and to-the-point you were able to make them. I do have some suggestions for ways we can make this even better - I've left those inline. Let me know what you think :)

drkrillo:
Thanks for the review! All changes made and ready for another look :)

Once this PR is merged, I'll be working on linking the terms to the first occurrence in other pages

s-makin (approved):
Thanks for making the changes, this all renders perfectly! All looks good to go :)

PR #731, opened 2026-06-12
document docker.io rootless mode

grayson-wolf:
Fixes #684

First draft, combines upstream installation guide with some of the tips section.


docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -593,6 +593,114 @@ $ docker logs web-server
...
+$ sudo systemctl disable --now docker.service docker.socket
+$ sudo rm /var/run/docker.sock
+```
+
+::::
+
+Check `/usr/bin` for the presence of `dockerd-rootless-setuptool.sh`. If it is not present, install it with
+
+```bash
+$ sudo apt-get install docker-ce-rootless-extras
```
athos-ribeiro: This is not a thing in Ubuntu. I don't think we want to document the installation of external packages.

athos-ribeiro (changes requested):
Thanks, Grayson!

Please check https://bugs.launchpad.net/ubuntu/+source/docker.io-app/+bug/2121905.

That bug, fixed in Resolute, introduced a couple scripts to allow setting up rootless docker.

The idea here is to play a bit with those, and document how users are supposed to set it up.

I do not believe we want to instruct users to install packages external to Ubuntu since we do not support those.

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,128 @@ docker logs web-server
 Error response from daemon: configured logging driver does not support reading
 ```
 
+## Rootless Docker
+
+Starting in Ubuntu 26.04 (package version `docker.io-app_29.1.3-0ubuntu4`), Docker supports running both containers and the Docker daemon itself as a non-root user. This is a useful way to provide extra security if it is not desirable to run the Docker daemon as root. As long as the prerequisites are all met, you do not even need root permissions to *install* Docker with rootless mode.
```
athos-ribeiro: > you do not even need root permissions to *install* Docker with rootless mode

This may sound misleading. Should it say that they do not need root to set the rootless mode up in that case?

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,128 @@ docker logs web-server
...
+
+If you are already running the rootful Docker daemon, disable it before proceeding:
+
+```bash
+$ sudo systemctl disable --now docker.service docker.socket
+$ sudo rm /var/run/docker.sock
+```
+::::
+
+Ubuntu installs the scripts related to rootless Docker in `/usr/share/docker.io/contrib`. This directory should be added to your `$PATH`:
```
athos-ribeiro: Based on the review of the script changes MP, I hope we can drop this step from here.

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,128 @@ docker logs web-server
...
+### Persisting the Rootless Daemon
+
+If you want to persist the rootless daemon across reboots and user logouts, enable the `docker` systemd user service and enable linger for your user:
+
+```bash
+$ systemctl --user enable docker
+$ loginctl enable-linger $(whoami)
+```
+
+### Running Rootless Docker inside Rootful Docker
```
athos-ribeiro: IMHO, we could drop this section.

If we we really want to keep a section like this (i.e., documenting specific docker images maintained by Docker Inc), this would need more explanation. For instance, what is docker:<version>-dind, how to use it, etc.

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,128 @@ docker logs web-server
...
+
+If you want to run rootless Docker inside a rootful Docker container, change your image to `docker:<version>-dind-rootless` instead of `docker:<version>-dind`:
+
+```bash
+$ docker run -d --name dind-rootless --privileged docker:<version>-dind-rootless
+```
+
+This image runs as UID 1000 by default, but `--privileged` is still required to disable seccomp, AppArmor, and mount masks.
+
+### Expose Docker API Socket
```
athos-ribeiro: Can we expand this?

How will exposing the socket help users? What are the use cases? Can we have an example in the end of the section?

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,128 @@ docker logs web-server
...
+```bash
+docker -H ssh://<REMOTEUSER>@<REMOTEHOST> run ...
+```
+
+### Expose Privileged Ports
+
+To expose ports less than 1024 (e.g 22 for SSH), set `CAP_NET_BIND_SERVICE` on the `rootlesskit` binary and restart the daemon:
+
+```bash
+$ sudo setcap cap_net_bind_service=ep $(which rootlesskit)
```
athos-ribeiro: It would be better to use the absolute path of the rootlesskit binary here to avoid granting such capabilities to some non-supported binary in the user path.

athos-ribeiro (commented):
Thanks, Grayson.

I added some comments inline.

docs/how-to/containers/docker-for-system-admins.md
grayson-wolf: Good catch, I clarified this.

docs/how-to/containers/docker-for-system-admins.md
grayson-wolf: The script in the MP has indeed been changed so this is no longer necessary, it's gone now.

docs/how-to/containers/docker-for-system-admins.md
grayson-wolf: I only kept this one in because it existed in the upstream docs and seemed marginally useful, but given that this is about specific upstream images I think I agree with you that this doesn't need to be here, it's dropped.

docs/how-to/containers/docker-for-system-admins.md
grayson-wolf: Done

docs/how-to/containers/docker-for-system-admins.md
grayson-wolf: The use case here is really just "I'm running rootless docker somewhere else and I want to do things on that box from some other box without manually sshing every time", the examples would just be running usual docker commands but they actually run remotely. I added a bit of extra detail here about why one might want to do this (and added some more detail in the SSH section about how to make it more transparent by setting up a context), let me know if you think even more might be needed

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
 Error response from daemon: configured logging driver does not support reading
 ```
 
+## Rootless Docker
+
+Starting in Ubuntu 26.04 (package version `docker.io-app_29.1.3-0ubuntu4`), Docker supports running both containers and the Docker daemon itself as a non-root user. This is a useful way to provide extra security if it is not desirable to run the Docker daemon as root. As long as the prerequisites are all met, you do not need root privileges to convert a rootful Docker installation into a rootless one.
+
+### Prerequisites
+
+You will need to install the following packages:
```
s-makin:
```suggestion
Install the following packages:
```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+:dir:
+sudo apt install rootlesskit uidmap slirp4netns fuse-overlayfs
+```
+
+- `rootlesskit`: A tool for running rootless containers.
+- `uidmap`: Provides `newuidmap` and `newgidmap` commands for managing UID and GID mappings.
+- `slirp4netns`: Required for user-mode networking in rootless containers.
+- `fuse-overlayfs`: User-space overlay filesystem for rootless containers.
+
+You will also need to ensure that the user in question has at least 65,536 subordinate UIDs and GIDs. To check this, consult the `/etc/subuid` and `/etc/subgid` files:
```
s-makin:
```suggestion
Ensure that the user in question has at least 65,536 subordinate UIDs and GIDs. To check this, consult the `/etc/subuid` and `/etc/subgid` files:
```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+```
+```{terminal}
+:copy:
+:user:
+:host:
+:dir:
+grep ^$(whoami): /etc/subgid
+```
+
+Each of these should output a line like
```
s-makin, line 840:
```suggestion
Each of these should output a line like

```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+```
```
s-makin:
```suggestion
```{terminal}
:output-only:
```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+yourusername:XXXXXX:65536
+```
+
+The second number is the number of subordinate UIDs or GIDs that the user has available. This value should be at least 65,536. f it is not, you will need to edit the `/etc/subuid` and `/etc/subgid` files to increase it.
```
s-makin:
```suggestion
The second number is the number of subordinate UIDs or GIDs that the user has available. This value should be at least 65,536. If it is not, edit the `/etc/subuid` and `/etc/subgid` files to increase it.
```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+
+```{terminal}
+:copy:
+:user:
+:host:
+:dir:
+/usr/share/docker.io/contrib/dockerd-rootless-setuptool.sh install
+```
+
+This command will also show help if any of the prerequisites are not met.
```
s-makin:
```suggestion
This command also shows help if any of the prerequisites are not met.
```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+
+After completing, the installation, you can confirm Docker is running in rootless mode with `docker info` and checking the context:
```
s-makin:
```suggestion
After completing the installation, confirm Docker is running in rootless mode with `docker info` and checking the context:

```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+```{terminal}
+:copy:
+:user:
+:host:
+:dir:
+docker info | grep Context
+```
+You should see the following line:
+```
+Context: rootless
```
s-makin:
```suggestion

Context: rootless
```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+```
+
+### Persisting the Rootless Daemon
```
s-makin:
```suggestion
### Persisting the rootless daemon
```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+```
+```{terminal}
+:copy:
+:user:
+:host:
+:dir:
+loginctl enable-linger $(whoami)
+```
+
+### Expose Docker API Socket
```
s-makin:
```suggestion
### Expose Docker API socket
```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+
+Exposing the Docker API Socket allows you to control the rootless Docker daemon from another machine. After connecting to a remote host, commands such as `docker ps` or `docker compose` will execute there. This could be useful if you, for instance, have a powerful server and a weak laptop, and you want to run containers on the server while controlling them from the laptop.
```
s-makin, line 917: This is great, a really nice feature to point out

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+
+#### Over SSH (Recommended)
```
s-makin:
```suggestion
#### Over SSH (recommended)
```

docs/how-to/containers/docker-for-system-admins.md
```diff
@@ -758,6 +758,196 @@ docker logs web-server
...
+:user:
+:host:
+:dir:
+DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS="-p 0.0.0.0:2376:2376/tcp" /usr/share/docker.io/contrib/docker-rootless.sh \
+   -H tcp://0.0.0.0:2376 \
+   --tlsverify --tlscacert=ca.pem --tlscert=cert.pem --tlskey=key.pem
+```
+
+
+### Expose Privileged Ports
```
s-makin:
```suggestion
### Expose privileged ports
```

s-makin (changes requested):
Mostly just formatting nits, nice work overall

docs/how-to/containers/docker-for-system-admins.md
athos-ribeiro: This is exactly what I had in mind. Thanks!

athos-ribeiro (approved):
Thanks for the changes. LGTM

s-makin (approved):
lgtm :)

PR #732, opened 2026-06-15
Convert admonition blocks to use colon fences

s-makin:
### Description

We have a mixture of backticks and colon fences enclosing our admonitions - here, we standardize on colon fences for all admonition blocks (note, warning, etc). Since admonitions can contain code blocks that use backticks, using backticks everywhere can lead to errors in how Sphinx renders them.

Having only one method for creating admonition blocks simplifies our guidance, and is in alignment with the Desktop docs which already only uses colon-fences.

### Checklist

---

### AI disclosure

I used Claude Opus 4.8 to handle the conversion - it also created a Python script (included in the `.github` folder


s-makin:
Looks like there was another change to the spelling checker, new words that were previously accepted (and some real words that it should be accepting) are being triggered. Fixed those up.

PR #734, opened 2026-06-15
Correction: SEV-SNP on EPYC started from Milan

alanbach:
### Description

The following one word change corrects the availability of SEV-SNP in EPYC to Milan. Current Rome is an incorrect statement.


s-makin (approved):
Thanks for the update! :)

PR #735, opened 2026-06-23
Update code blocks

### LINKED ISSUE #102: Style consistency when showing commands and their results
We have a mix of:

a) show command, and in a separate code block, show the output. In this case, the command part is shown WITHOUT a leading shell prompt (i.e., like `# ` or `$ `);

b) show command and output in the same code block. In this case, the the command part is shown WITH a leading shell prompt.

The argument in favor of (a) is that it's easier for people to copy the command with one click, whereas in (b) they have to select the command exactly, being careful to exclude the leading shell prompt. Going in favor of (b), though, is that it's better presented and is generally a shorter form, which doesn't need an introduction ("here is the command:", and "here is the result:").

My personal preference, visually, is (a), but it lacks a clear distinction between what is that the user has to type, and what is the result, whereas in (b) that is not an issue.

s-makin:
### Description

This PR updates the majority of code blocks in the Server documentation to use the Sphinx `{terminal}` directive. It also adds usage instructions to the contributing docs.

Where a former code block clearly referred to a command (input/output/both) this has been converted to better represent what a user will see at the terminal.

Actual code (e.g. scripts, yaml config, json, etc) is still code blocks, with the language correctly marked.

Config files that I could definitively identify as config files are marked as "ini".

Everything else is "text", to suppress the weird syntax colouring you get with the default.

#### Why
- To improve consistency between Ubuntu documentation sets
- To resolve the question of how to show command prompts in a code block (or not)
- To simplify/standardize how we format commands

#### Known issues with the terminal directive

(for which I've filed issues upstream)
- It doesn't allow arbitrary prompts: pages documenting Windows-related content, and any code blocks where commands are run as root are the same as before
- It doesn't allow "empty" lines to demonstrate moving to a different machine, so former code blocks that show a transition to a different machine have this included in the output of the previous command
- Continuation characters (`>`) in multiline prompts are carried over when copying

---

### Related Issue

- Fixes: #102 

### Checklist

---


cpaelzer (approved):
Hi Sally,
I checked a few pages to be sure, so far things look fine and you have my +1

Let me ask one question for clarification:
1. if we have only the command -> got converted
2. if we had command and output split -> only the command got converted
3. if we had the command and output in one -> got converted

Is that as it is intended?

For example seen on https://canonical-ubuntu-server-documentation--735.com.readthedocs.build/how-to/networking/serve-ntp-with-chrony/ it has all variants of it.

s-makin:
My schema was:

>     1. if we have only the command -> got converted

These got converted (if they are not excluded by being root or powershell, because of limitations of the extension)
 
>     2. if we had command and output split -> only the command got converted

They should all be converted - in many cases I combined them into a single terminal block, unless the discussion between original command and output made sense in between, I kept them separate.

>     3. if we had the command and output in one -> got converted

They should all be converted
 
> For example seen on https://canonical-ubuntu-server-documentation--735.com.readthedocs.build/how-to/networking/serve-ntp-with-chrony/ it has all variants of it.

The blocks I didn't convert (throughout) are either the contents of conf files, logs, or actual code. 

In this case, the blocks with `chronyc>` didn't get converted because we can't use arbitrary promps, but I do see there are some output blocks that should have been converted that I didn't notice.

PR #739, opened 2026-06-25
debuginfod: add bug report url and simplify buildid query

TheJJ:
- add bug report url
- directly query build with readelf instead of displaying all notes.


s-makin (approved):
lgtm :)

PR #745, opened 2026-07-02
unattended-upgrades: add section to cover the needrestart

### LINKED ISSUE #744: [Feedback]: Automatic updates: do not auto-restart critical services
### Page URL

https://ubuntu.com/server/docs/how-to/software/automatic-updates/

### Select an option

- [x] I couldn't find what I was looking for

### Issue Description

It could be useful to say that automatic updates may cause critical services to be auto-restated (e.g. `systemctl restart whatever-critical-service.service` - and in production this may be very problematic. But this can be configured documenting yourself about the `needrestart` package. Source:

https://askubuntu.com/a/1530298/263638

hector-cao:
Fixes #744

### DIFF
--- a/docs/how-to/software/automatic-updates.md
+++ b/docs/how-to/software/automatic-updates.md
@@ -351,7 +351,7 @@ Below are the logs of an `unattended-upgrades` run that started at 20:43. The to
 
 Many updates do not require a full reboot, but do require services that use the updated files to be restarted before the change takes effect. For example, updating a shared library means every running service linked against it keeps using the old version in memory until it is restarted.
 
-This is handled by the {manpage}`needrestart(1)` package, which `unattended-upgrades` calls after applying updates. Starting with Ubuntu 24.04 LTS (see [discourse post](https://discourse.ubuntu.com/t/needrestart-changes-in-ubuntu-24-04-service-restarts/44671)), `needrestart` restarts the affected services automatically by default.
+This is handled by the {manpage}`needrestart(1)` package, which `unattended-upgrades` calls after applying updates. [Starting with Ubuntu 24.04 LTS](https://discourse.ubuntu.com/t/needrestart-changes-in-ubuntu-24-04-service-restarts/44671), `needrestart` restarts the affected services automatically by default.
 
 It does not restart everything, though. It ships the default configuration file `/etc/needrestart/needrestart.conf` with a pre-set list of known cases that are never restarted automatically (such as the display manager), so only the services outside that list are affected.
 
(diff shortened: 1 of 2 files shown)


cpaelzer:
Mostly LGTM - thanks!

I wonder if it would be worth to mention that needrestart has an internal list of things it would not restart? It is not that "all but what you exclude" are restarted. Instead /etc/needrestart/needrestart.conf starts with some known cases pre-set.

Since user changes to the conf.d are so much better along upgrades, how about only suggesting that in the text when explaining how to adapt it?

hector-cao:
Thanks @cpaelzer ,

I tried to improve it based on your comment, let me know if it is better or worse.

docs/how-to/software/automatic-updates.md
```diff
@@ -347,6 +347,27 @@ Below are the logs of an `unattended-upgrades` run that started at 20:43. The to
 2025-03-13 20:43:40,207 WARNING Shutdown msg: b"Reboot scheduled for Thu 2025-03-13 20:45:00 UTC, use 'shutdown -c' to cancel."
 ```
 
+## Service restarts
+
+Many updates do not require a full reboot, but do require services that use the updated files to be restarted before the change takes effect. For example, updating a shared library means every running service linked against it keeps using the old version in memory until it is restarted.
+
+This is handled by the {manpage}`needrestart(1)` package, which `unattended-upgrades` calls after applying updates. Starting with Ubuntu 24.04 LTS (see [discourse post](https://discourse.ubuntu.com/t/needrestart-changes-in-ubuntu-24-04-service-restarts/44671)), `needrestart` restarts the affected services automatically by default.
```
s-makin:
```suggestion
This is handled by the {manpage}`needrestart(1)` package, which `unattended-upgrades` calls after applying updates. [Starting with Ubuntu 24.04 LTS](https://discourse.ubuntu.com/t/needrestart-changes-in-ubuntu-24-04-service-restarts/44671), `needrestart` restarts the affected services automatically by default.
```

s-makin (changes requested):
Generally lgtm, just one suggestion. The CI is picking up a spelling error on "boolean" but that's a valid term - could you please add that to the `.custom_wordlist` file to clear that error?

hector-cao:
> Generally lgtm, just one suggestion. The CI is picking up a spelling error on "boolean" but that's a valid term - could you please add that to the `.custom_wordlist` file to clear that error?

thanks ! done

s-makin (approved):
Thanks for the quick fixes - lgtm :) I'll hold off on merging until you hear back from @cpaelzer

utkarsh2102 (approved):

PR #747, opened 2026-07-06
add migration doc from wiki contents

hector-cao:
Migrate the Wiki content for virtualization live migration feature with QEMU/libvirt:

https://wiki.ubuntu.com/QemuKVMMigration#Use_Cases


cpaelzer:
hector and I deeply reviewed this, great work but much to change because of my old mistakes!

hector-cao:
Hey @cpaelzer , I did several improvements following our discussion on that MR:
- separate live migration into a separate page
- improve the existing machine types page
- added the instruction on how to compute the lowest CPU denominator
- some wording and nitpicks

renanrodrigo:
<intruder comment>
Shouldn't this now be `virtualiZation` (and other terms revised for the same) now that we default to EN_US ?
</me flies away>

docs/redirects.txt
```diff
@@ -98,7 +98,7 @@ vm-tools-in-the-ubuntu-space explanation/virtualisation/vm-tools-in-the-ubuntu-s
 
 using-qemu-for-microvms explanation/virtualisation/qemu-microvm/
 
-upgrading-the-machine-type-of-your-vm explanation/virtualisation/upgrading-the-machine-type-of-your-vm/
```
cpaelzer, line 101: I think you need to keep the old ref (kinda forever IIU @s-makin correctly) but point it to the new position. And since "live-migration" never existed there is no need to redirect this - that would only be to catch the old high level entry to point it to where it is now sorted in.

since ‎docs/explanation/virtualisation/upgrading-the-machine-type-of-your-vm.md goes away and is consumed/replaced by docs/explanation/virtualisation/qemu-machine-types.md‎ I'd have more expected something like

```
explanation/virtualisation/upgrading-the-machine-type-of-your-vm explanation/virtualisation/qemu-machine-types
```
rather than anything else, but I'm unsure.
Sorry, those redirects can be confusing - maybe Sally can help more

docs/index.md
```diff
@@ -52,7 +52,7 @@ Install, update, upgrade, and troubleshoot the software packages running on your
 
 * **Managing software**: {ref}`Package management <package-management>` • {ref}`etckeeper <install-etckeeper>` • {ref}`Get older package versions using snapshots <snapshot-service>`
 * **Update considerations**: {ref}`Automatic updates <automatic-updates>` • {ref}`Phased updates <about-apt-upgrade-and-phased-updates>` • {ref}`Testing updates in advance <advance-testing-of-updates-in-best-practice-server-deployments>`
-* **Upgrade Ubuntu version**: {ref}`Release upgrades <upgrade-your-release>` • {ref}`Upgrade a virtual machine <upgrading-the-machine-type-of-your-vm>`
+* **Upgrade Ubuntu version**: {ref}`Release upgrades <upgrade-your-release>`
```
cpaelzer, line 55: ack, dropped here to both end up in explanation as shown above
Nothing to change, mostly a reminder for myself

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
+---
+myst:
+  html_meta:
+    description: Understand QEMU machine types on Ubuntu, and how to upgrade the machine type of an existing guest.
+---
+
+(qemu-machine-types)=
+# QEMU Machine types
+
+The QEMU machine type defines the exact virtual hardware blueprint (chipset, PCI slots, ACPI tables) of a virtual machine. This allows QEMU to offer to virtual machines a well-defined and stabilized foundation they can run on so that they are not impacted by changes that happen under the hood on the virtualisation software stack at machine upgrade or guests migration.
```
cpaelzer:
```suggestion
The QEMU machine type defines the exact virtual hardware blueprint (chipset, bus architecture, and version-locked features) of a virtual machine. This allows QEMU to offer to virtual machines a well-defined and stable environment they can run not impacted by changes that happen under the hood like software upgrades or migrations to other systems.
```

The protection is to stay stable in behavior when code changes.
I also have been pondering about the list in the blueprint to make a proposal that (to me) seems better.

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+
+However, when the QEMU software is upgraded, machine types can evolve and get new features that might be incompatible with the migrated virtual machine. The solution is to version the machine types. The following is the list of Q35 machine types available in QEMU 10.2:
```
cpaelzer:
```suggestion
However, when the QEMU software is upgraded, it might get new features and behavior that would be incompatible with the current definition virtual machine. But since the purpose of these types is to lock the feature set, they can't be changed - the solution is to version the machine types. The following is the list of Q35 machine types available in QEMU 10.2:
```

Feel free to dislike it, but I think it is an improvement to the former

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+pc-i440fx-resolute   Ubuntu 26.04 PC (i440FX + PIIX, 1996) (default)
+ubuntu-q35           Ubuntu 26.04 PC (Q35 + ICH9, 2009) (alias of pc-q35-resolute)
+pc-q35-resolute      Ubuntu 26.04 PC (Q35 + ICH9, 2009)
+```
+
+`pc-q35-resolute` and `pc-i440fx-resolute` are forked respectively from `pc-q35-10.2` and `pc-i440fx-10.2` that are the latest x86 upstream machine types since QEMU 10.2 is the version for Ubuntu Resolute LTS. `pc-i440fx-resolute` is defined as the default machine type for every virtual machine and the `ubuntu` alias now points to it.
+
+These distro-specific machine types allow Ubuntu to support users who have deployed VMs on Ubuntu LTSes to be able to live migrate their VMs to the next LTS dealing with various changes we might introduce into an existing Ubuntu release through the SRU process. Whenever we need to make a change that modifies an existing machine type, we can make a "point-release" element in the Ubuntu names. This is not tied to a usual Ubuntu LTS point release, but to
+anything introducing a delta to the machine type / `vmstate`. Similar to CentOS/RHEL, we want:
```
cpaelzer:
```suggestion
anything introducing a delta to the machine type / `vmstate`. The format generally follows:
```

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+
+```text
+$MACHINE_TYPE-$DISTRO-RELEASE[-v*]
+```
+
+This is an example of "point-release" machine types that have been added to address the [arch-capabilities bug](https://bugs.launchpad.net/bugs/2131822) in Noble:
```
cpaelzer:
```suggestion
The following is an example of an update that required machine types versioning to track a change in behavior when addressing a bug in handling [arch-capabilities bug](https://bugs.launchpad.net/bugs/2131822) in Noble:
```

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+
+```text
+pc-i440fx-questing-v2 Ubuntu 25.10 PC v2 (i440FX + PIIX, + 10.1 machine, 1996)
+pc-i440fx-noble-v2   Ubuntu 24.04 PC v2 (i440FX + PIIX, arch-caps fix, 1996)
+pc-q35-noble-v2      Ubuntu 24.04 PC v2 (Q35 + ICH9, arch-caps fix, 2009)
+```
+
+## Machine type handling summary
```
cpaelzer: I'd ask you to drop all from "Machine types are handled by:" to "The default, if no machine type is specified, always points to the latest distribution-specific machine type."
You have explained all that above and doing so again is only redundancy and distracting.

After removal change the header "Machine type handling summary" => "Machine type availability"
The then remaining elements talk about why we keep upstream names and if/when we cleanup which fits the title.

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+  - ppc64el: `pseries-resolute`
+  - arm64: `virt-resolute`
+- Feature backports add a `-v%d` suffix to the affected types.
+  - To avoid a proliferation of these types, such changes should be done carefully.
+- The default, if no machine type is specified, always points to the latest
+  distribution-specific machine type.
+- We do not drop upstream types; they are provided as-is without further
+  guarantees.
+  - Cross-vendor/downstream migrations might work for upstream types, but are
+    considered unsupported. This has always been the case, but package documentation
```
cpaelzer: We discussed that in 1:1, drop all of "This has always been the case, but package documentation might need updating to reflect this."

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+    might need updating to reflect this.
+- Cleanup matches the usual supported distribution upgrade paths.
```
cpaelzer, line 76: We also discussed that Cleanup section to be silly nowadays and was an overlay aggressive past intention but never acted on, we had a better version that was a draft like:

```
=> The supported migration paths match the supported Ubuntu upgrade paths. So a release can always migrate to the subsequent release, and furthermore an LTS can migrate to the following LTS+1. In addition to that an LTS can be expected to also safely migrate to LTS+2. Other paths might work as well - in fact, many do - but only those matching upgrade paths are officially considered supported.
```

And we had this which is a nice "why do they even exist" to go where you explain the ubuntu themed names above:
```
=> types are for recognition as “ubuntu” and for avoiding the misbelief that they could neatly move elsewhere while we can’t give any guarantees. Furthermore redhat paved the way and they have distro named types as well
```

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+  - Drop former non-LTS release definitions after the next LTS.
+  - Drop former LTS release definitions when out of support.
+
+## Upgrading the machine type of your VM
+
+Upgrading the machine type of a virtual machine (VM) can be thought of in the same way as buying (virtual) hardware of the same spec but with a newer release date. Whereas to upgrade a physical machine you might buy an improved CPU, more RAM, or increased storage, with a virtual machine you can change the configuration to achieve the same results.
```
cpaelzer:
```suggestion
Think about upgrading a physical machine:
* You might buy more CPUs and RAM, or increased storage - for a virtual machine you can to that by a change to its configuration.
* But you'd usually also buy a newer generation of board, chipset and so on - in a virtual machine the equivalent to that is using a newer machine type
```

I might have rewritten this too often, hope you like it

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Why should you do this for a VM?
```
cpaelzer: Drop the whole "Why should you do this for a VM?" section please.
It is a lot "bla bla" and barely and holds up to a hard question. It could be super extended and then make sense, but isn't important to the story told. Remove it please. The updated suggested text above it making it clear what it is, people are used to think in real HW and that works IMHO much better.

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+Change this to the value you want. If you need to check what machine types are available via the `kvm -M ?` command first, then note that while upstream types are provided for convenience, only Ubuntu types are supported. There you can also see what the current default would be, as in this example:
+
+```{terminal}
+:copy:
+:user:
+:host:
+:dir:
+kvm -M ?
+
+pc-i440fx-xenial       Ubuntu 16.04 PC (i440FX + PIIX, 1996) (default)
```
cpaelzer: While touching it - Maybe bump all the old examples to Noble+Resolute examples?

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+virsh list --inactive
+```
+
+#### Edit the guest definition
+
+Once the machine is listed as "shut off", you can then edit the machine definition and find the type in the `type` tag given at the machine attribute.
+
+```bash
+virsh edit <your_machine>
+<type arch='x86_64' machine='pc-i440fx-bionic'>hvm</type>
```
cpaelzer: Nowadays virt-xml might do that more comfortably - see https://manpages.ubuntu.com/manpages/resolute/man1/virt-xml.1.html#boot and try if that works well for ubuntu types.

If it works see if the description would seem better with this and refer to the man page for more.

If it does not work that might be a patch that we need to track as bug anyway (low prio but valid).

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,104 @@
...
+
+- **Versioned machine types**, which keep the guest's virtual hardware layout stable across QEMU versions.
+- **CPU model + named features**, which allows the CPU to be expressed as a baseline model name +/- named features.
+
+## Versioned machine types
+
+The QEMU machine type defines the exact virtual hardware blueprint (chipset, PCI slots, ACPI tables) of a virtual machine. This allows QEMU to offer to virtual machines a well-defined and stabilized foundation they can run on so that they are not impacted by changes that happen under the hood on the virtualisation software stack. The machine types happen to be very useful to ensure live migration succeeds, especially inside heterogeneous infrastructure.
```
cpaelzer:
```suggestion
```

Drop this as it is better explained in the extra page now.

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,104 @@
...
+
+When you live-migrate a VM, you are taking a snapshot of a running system's RAM and CPU state and sending it over the wire. The destination host must have a virtual hardware environment that matches that state down to the very last byte; this is possible by keeping the machine type unchanged. However, when the QEMU software is upgraded, machine types can evolve and get new features that might be incompatible with the migrated virtual machine. The solution is to version the machine types by creating a new version of the machine type for each supported QEMU version.
```
cpaelzer:
```suggestion
When you live-migrate a VM, you are taking a snapshot of a running system's RAM and CPU state and sending it over the wire. The destination host must have a virtual hardware environment that matches that state exactly; this is possible due to both sides of the migration agreeing on the same QEMU machine type.
```

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,104 @@
...
+
+For more detail on Ubuntu machine types and how to upgrade the machine type of an existing guest, see {ref}`qemu-machine-types`.
+
+## CPU baseline and named features
+
+Versioned machine types keep the virtual platform stable, but they do not control the CPU that the guest sees. The CPU is exposed separately, and it is the second mechanism that makes migration across heterogeneous hosts possible.
+
+Libvirt and QEMU allow the CPU to be expressed in ways that ease migration between hosts in heterogeneous infrastructures. While a libvirt VM configured with `host-passthrough` or `max` (the equivalent QEMU `-cpu host` and `-cpu max`) takes the most out of the hardware and kernel hypervisor, it exposes the guests to all available host capabilities. Therefore, migration to another host is much less transparent to the guests (by exposing new capabilities or deprecating existing features), and this can break the guests after migration. Instead of specifying the CPU capabilities in such manner, a CPU can be expressed as a baseline CPU model +/- a list of features. In QEMU, it comes down to this example `-cpu` argument:
```
cpaelzer:
```suggestion
Libvirt and QEMU allow the CPU to be expressed in ways that ease migration between hosts in heterogeneous infrastructures. While a libvirt VM configured with `host-passthrough` or `maximum` (the equivalent QEMU `-cpu host` and `-cpu max`) gets most out of the hardware, it exposes the guests to all available host capabilities. Therefore, migration to another host is much more restricted, because it needs to be able to present the very same environment to the guest. To do so it needs a clear definition that helps to ensure both sides of the migration can present the same. Therefore it is preferred to specifying the CPU capabilities as a baseline CPU model +/- a list of features to ensure compatibility. In QEMU, it comes down to this example `-cpu` argument:
```

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,104 @@
...
+
+```text
+-cpu EPYC-Turin,x2apic=on,tsc-deadline=on,hypervisor=on,tsc-adjust=on,
+ spec-ctrl=on,stibp=on,flush-l1d=on,ssbd=on,cmp-legacy=on,virt-ssbd=on,
+ tsa-sq-no=on,tsa-l1-no=on,pcid=off,la57=off
+```
+
+You can ask libvirt to build the CPU for you by using the `host-model` CPU mode:
```
cpaelzer:
```suggestion
Or in a level higher, you can ask libvirt to build the CPU for you by using the `host-model` CPU mode.
This will detect the current CPU with all its features, but automatically expresses it as baseline CPU model +/- a list of features.
```

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,104 @@
...
+
+```xml
+...
+  <cpu mode='host-model' check='partial'/>
+...
+```
+
+or provide the exact CPU specification:
```
cpaelzer:
```suggestion
or you might prefer even more control and provide the exact CPU specification:
```

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,104 @@
...
+    <feature policy='require' name='x2apic'/>
+    <feature policy='require' name='tsc-deadline'/>
+    ...
+    <feature policy='require' name='tsa-l1-no'/>
+    <feature policy='disable' name='pcid'/>
+    <feature policy='disable' name='la57'/>
+...
+```
+
+While the first approach seems easier and more appealing, it is not suitable for heterogeneous infrastructure. To retain full control over migratability across every host in the migration pool, the second approach is better. Explicitly specifying the CPU definition advertises to the guest only the CPU features that are guaranteed to exist on every host in the pool, rather than everything the current host happens to offer. This creates a stable "lowest common denominator" CPU that looks identical on both the source and the destination, so the guest never comes to rely on a feature that could disappear after a migration.
```
cpaelzer:
```suggestion
While the first approach of just using `host-passthrough` and even `host-model` seems easier and more appealing, it is not suitable for heterogeneous infrastructure. To retain full control over migratability across every host in the migration pool a CPU should be modeled.
You can check the capabilities of every host in your pool and define the CPU definition to only advertise CPU features to the guest that are guaranteed to exist on every host in the pool.
This creates a stable "lowest common denominator" CPU that looks identical on both the source and the destination, so the guest never comes to rely on a feature that could disappear after a migration.
```

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+### Determining the lowest common CPU denominator
+
+
+Finding the lowest denominator seems to be a complex task. Fortunately, libvirt provides a command to compute it (see {manpage}`virsh(1)` and its `hypervisor-cpu-baseline` command) by taking the CPU definitions of every host in the pool and returning a single baseline CPU that all of them can support.
```
cpaelzer, line 76: Thank you for adding this section

cpaelzer (changes requested):
To be clear of the outside viewer, 50% of what I criticize here is my own text that I wrote a decade ago to the wiki. The rest is juts helpful guidance and suggestions to make this the most helpful for a future reader of it.

docs/explanation/virtualisation/live-migration.md
hector-cao: I struggle to be convinced by the sentence: "Instead of `max` to ensure compatibility it is therefore preferred to specifying the CPU capabilities ...", can you confirm that this is the text you like to propose, if yes, please help me there

docs/explanation/virtualisation/live-migration.md
hector-cao: When I mention "first approach", it is about using the "host-model" , not "host-passthrough", host-passthrugh has been definitely discarded earlier for migration consideration

hector-cao:
> Shouldn't this now be `virtualiZation` (and other terms revised for the same) now that we default to EN_US ?

thanks. Done

docs/explanation/virtualisation/live-migration.md
cpaelzer: Yes, that is what I'd suggest even when re-reading.
Maybe accept it as is and we wait for Sally's review next which will tear both our sentences apart for the greater good anyway?

docs/explanation/virtualisation/live-migration.md
cpaelzer: That was not clear, but I'd not put host-model too far away.
Could is be ?

```
While the first approach of just using `host-passthrough` and even `host-model` seem easier and more appealing, it is not suitable for heterogeneous infrastructure. To retain full control over migratability across every host in the migration pool a CPU should be modeled to a common denominator.
```

docs/explanation/virtualisation/live-migration.md
hector-cao: It is more about the technical meaning,
what do you mean by `max` to ensure compatibility ? which compatibility do you have in mind ?

docs/explanation/virtualisation/live-migration.md
hector-cao: yes, I modified your suggestion and applied it. Thanks

docs/redirects.txt
cpaelzer: Expected to change the existing line to something like
`upgrading-the-machine-type-of-your-vm docs/explanation/virtualisation/qemu-machine-types/`

Also, why are we adding "live-migration" line, nothing moves, could be dropped

cpaelzer: the above is for the old source to the new target
and for the new move something like
`docs/explanation/virtualisation/upgrading-the-machine-type-of-your-vm/ docs/explanation/virtualisation/qemu-machine-types/`

hector-cao: Thanks, I updated the MR

cpaelzer (approved):
Thanks for all the adaptations and discussions - content is good for me now.

cpaelzer:
@s-makin Hector and I are done iterating on the structure and content, could you please ensure we didn't leave awkward sentences around and merge once ok?

s-makin:
> @s-makin Hector and I are done iterating on the structure and content, could you please ensure we didn't leave awkward sentences around and merge once ok?

I have a gap today where I can take a look :eyes:

docs/explanation/virtualisation/live-migration.md
s-makin: What does "over the wire" mean? This feels quite colloquial

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,107 @@
...
+
+- **Versioned machine types**, which keep the guest's virtual hardware layout stable across QEMU versions.
+- **CPU model + named features**, which allows the CPU to be expressed as a baseline model name +/- named features.
+
+## Versioned machine types
+
+The QEMU machine type defines the exact virtual hardware blueprint (chipset, PCI slots, ACPI tables) of a virtual machine. This allows QEMU to offer to virtual machines a well-defined and stabilized foundation they can run on so that they are not impacted by changes that happen under the hood on the virtualization software stack. The machine types happen to be very useful to ensure live migration succeeds, especially inside heterogeneous infrastructure.
```
s-makin:
```suggestion
The QEMU machine type defines the exact virtual hardware blueprint (chipset, PCI slots, ACPI tables) of a virtual machine. This allows QEMU to offer VMs a well-defined and stabilized foundation they can run on, so they are not impacted by changes that happen "under the hood" on the virtualization software stack. The machine types are useful to ensure live migration succeeds, especially inside heterogeneous infrastructure.
```

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,107 @@
...
+
+When you live-migrate a VM, you are taking a snapshot of a running system's RAM and CPU state and sending it over the wire. The destination host must have a virtual hardware environment that matches that state down to the very last byte; this is possible by keeping the machine type unchanged. However, when the QEMU software is upgraded, machine types can evolve and get new features that might be incompatible with the migrated virtual machine. The solution is to version the machine types by creating a new version of the machine type for each supported QEMU version.
+
+For more detail on Ubuntu machine types and how to upgrade the machine type of an existing guest, see {ref}`qemu-machine-types`.
+
+## CPU baseline and named features
+
+Versioned machine types keep the virtual platform stable, but they do not control the CPU that the guest sees. The CPU is exposed separately, and it is the second mechanism that makes migration across heterogeneous hosts possible.
```
s-makin: Unclear what you mean by "the second mechanism" (and by contrast, which one is the first)

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,107 @@
...
+
+Libvirt and QEMU allow the CPU to be expressed in ways that ease migration between hosts in heterogeneous infrastructures. While a libvirt VM configured with `host-passthrough` or `maximum` (the equivalent QEMU `-cpu host` and `-cpu max`) gets most out of the hardware, it exposes the guests to all available host capabilities. Therefore, migration to another host is much more restricted, because it needs to be able to present the very same environment to the guest. To do so it needs a clear definition that helps to ensure both sides of the migration can present the same. Therefore it is preferred to specifying the CPU capabilities as a baseline CPU model +/- a list of features to ensure compatibility. In QEMU, it comes down to this example `-cpu` argument:
```
s-makin:
```suggestion
Libvirt and QEMU allow the CPU to be expressed in ways that ease migration between hosts in heterogeneous infrastructures. While a libvirt VM configured with `host-passthrough` or `maximum` (the equivalents in QEMU are `-cpu host` and `-cpu max` respectively) gets the most out of the hardware, it exposes the guests to all available host capabilities. Therefore, migration to another host is much more restricted, because it needs to be able to present the very same environment to the guest. To do so it needs a clear definition that helps to ensure both sides of the migration can present the same. Therefore it is preferred to specifying the CPU capabilities as a baseline CPU model +/- a list of features to ensure compatibility. In QEMU, it comes down to this example `-cpu` argument:
```

s-makin: "gets the most out of the hardware" - gets the most what?

s-makin: "both sides of the migration can present the same" - the same what? Also, this paragraph feels ike a few points are being repeated a bit. Is there room for us to restructure a bit to avoid repetition?

s-makin: "Therefore it is preferred to specifying the..." -> "It is best to specify..."

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,107 @@
...
+
+```text
+-cpu EPYC-Turin,x2apic=on,tsc-deadline=on,hypervisor=on,tsc-adjust=on,
+ spec-ctrl=on,stibp=on,flush-l1d=on,ssbd=on,cmp-legacy=on,virt-ssbd=on,
+ tsa-sq-no=on,tsa-l1-no=on,pcid=off,la57=off
+```
+
+Or in a level higher, you can ask libvirt to build the CPU for you by using the `host-model` CPU mode.
```
s-makin: "in a level higher" - in what parameter space are we talking here? Level higher than what?

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,107 @@
...
+This will detect the current CPU with all its features, but automatically expresses it as baseline CPU model +/- a list of features.
```
s-makin:
```suggestion
This detects the current CPU with all its features, automatically expressed as baseline CPU model +/- a list of features.
```

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,107 @@
...
+
+```xml
+...
+  <cpu mode='host-model' check='partial'/>
+...
+```
+
+or you might prefer even more control and provide the exact CPU specification:
```
s-makin:
```suggestion
If you prefer more control, you can provide the exact CPU specification:
```

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,107 @@
...
+    <feature policy='require' name='x2apic'/>
+    <feature policy='require' name='tsc-deadline'/>
+    ...
+    <feature policy='require' name='tsa-l1-no'/>
+    <feature policy='disable' name='pcid'/>
+    <feature policy='disable' name='la57'/>
+...
+```
+
+While the first approach of just using `host-passthrough` and even `host-model` seems easier and more appealing, it is not suitable for heterogeneous infrastructure. To retain full control over migratability across every host in the migration pool a CPU should be modeled.
```
s-makin: I think it's more helpful for readers if we tell them right from the start which method they should follow so they can skip to the correct section, rather than telling them afterwards that a particular method isn't suitable for certain cases. Otherwise they might waste their time and effort trying something they shouldn't be.

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,107 @@
...
+
+
+### Determining the lowest common CPU denominator
+
+
+Finding the lowest denominator seems to be a complex task. Fortunately, libvirt provides a command to compute it (see {manpage}`virsh(1)` and its `hypervisor-cpu-baseline` command) by taking the CPU definitions of every host in the pool and returning a single baseline CPU that all of them can support.
+
+First, collect the host CPU definition from each host in the pool. The `<cpu>` block reported by `virsh capabilities` describes what the host CPU offers:
+
+```text
```
s-makin, line 80: Is this a command? If so, can we render it as a terminal block?

s-makin: (same comment for any other command blocks)

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,107 @@
...
+virsh capabilities > host1-caps.xml
+```
+
+Then feed all the files to `hypervisor-cpu-baseline`. Libvirt extracts the `<cpu>` element from each file and computes the baseline:
+
+```text
+virsh hypervisor-cpu-baseline --migratable host1-caps.xml host2-caps.xml
+```
+
+The `--migratable` option restricts the result to features that are safe to migrate, excluding those that cannot be preserved across a migration. The command prints a `<cpu>` block similar to this:
```
s-makin, line 90: What happens if there are features that can't be preserved? what should the user do if they want those features?

docs/explanation/virtualisation/live-migration.md
```diff
@@ -0,0 +1,107 @@
...
+<cpu mode='custom' match='exact'>
+  <model fallback='forbid'>EPYC-Turin</model>
+  <vendor>AMD</vendor>
+  <feature policy='require' name='x2apic'/>
+  <feature policy='require' name='tsc-deadline'/>
+  ...
+</cpu>
+```
+
+This block is the lowest common denominator CPU for the pool. Drop it into the domain XML of the VMs you intend to migrate, so that these VMs see the same CPU regardless of which host they run on.
```
s-makin: What do we mean by "drop it"? This might be confusing because it conflicts with the commonly used meaning being to discard something

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
+---
+myst:
+  html_meta:
+    description: Understand QEMU machine types on Ubuntu, and how to upgrade the machine type of an existing guest.
+---
+
+(qemu-machine-types)=
+# QEMU Machine types
+
+The QEMU machine type defines the exact virtual hardware blueprint (chipset, bus architecture, and version-locked features) of a virtual machine. This allows QEMU to offer to virtual machines a well-defined and stable environment they can run not impacted by changes that happen under the hood like software upgrades or migrations to other systems.
```
s-makin:
```suggestion
The QEMU machine type defines the exact virtual hardware blueprint (chipset, bus architecture, and version-locked features) of a virtual machine. This allows QEMU to offer VMs a well-defined and stable environment they can run without being impacted by changes that happen "under the hood", such as software upgrades or migrations to other systems.
```

s-makin: do we want to link to the live migration page here?

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+
+However, when the QEMU software is upgraded, it might get new features and behavior that would be incompatible with the current definition virtual machine. But since the purpose of these types is to lock the feature set, they can't be changed - the solution is to version the machine types. The following is the list of Q35 machine types available in QEMU 10.2:
```
s-makin:
```suggestion
However, when the QEMU software is upgraded, it might get new features and behavior that would be incompatible with the current definition of a given virtual machine. Since the purpose of these machine types is to lock the feature set, they can't be changed - so the solution is to version the machine types. The following is the list of Q35 machine types available in QEMU 10.2:
```

s-makin: Unclear what Ubuntu versions this is available for. Also, do we want to provide a table here that we need to maintain? Is there a QEMU official page we can link to instead?

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+pc-q35-8.2           Standard PC (Q35 + ICH9, 2009)
+pc-q35-9.0           Standard PC (Q35 + ICH9, 2009)
+pc-q35-9.1           Standard PC (Q35 + ICH9, 2009)
+pc-q35-9.2           Standard PC (Q35 + ICH9, 2009)
+pc-q35-10.0          Standard PC (Q35 + ICH9, 2009)
+pc-q35-10.1          Standard PC (Q35 + ICH9, 2009)
+pc-q35-10.2          Standard PC (Q35 + ICH9, 2009)
+```
+
+## Ubuntu specific machine types
```
s-makin:
```suggestion
## Ubuntu-specific machine types
```

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+
+```text
+root@resolute:~# qemu-system-x86_64 -machine help | grep resolute
+ubuntu               Ubuntu 26.04 PC (i440FX + PIIX, 1996) (alias of pc-i440fx-resolute)
+pc-i440fx-resolute   Ubuntu 26.04 PC (i440FX + PIIX, 1996) (default)
+ubuntu-q35           Ubuntu 26.04 PC (Q35 + ICH9, 2009) (alias of pc-q35-resolute)
+pc-q35-resolute      Ubuntu 26.04 PC (Q35 + ICH9, 2009)
+```
+
+`pc-q35-resolute` and `pc-i440fx-resolute` are forked respectively from `pc-q35-10.2` and `pc-i440fx-10.2` that are the latest x86 upstream machine types since QEMU 10.2 is the version for Ubuntu Resolute LTS. `pc-i440fx-resolute` is defined as the default machine type for every virtual machine and the `ubuntu` alias now points to it.
```
s-makin:
```suggestion
`pc-q35-resolute` and `pc-i440fx-resolute` are forked respectively from `pc-q35-10.2` and `pc-i440fx-10.2`, which are the latest x86 upstream machine types, since QEMU 10.2 is the version for Ubuntu Resolute LTS. `pc-i440fx-resolute` is defined as the default machine type for every virtual machine and the `ubuntu` alias now points to it.
```

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+
+These distro-specific machine types allow Ubuntu to support users who have deployed VMs on Ubuntu LTSes to be able to live migrate their VMs to the next LTS dealing with various changes we might introduce into an existing Ubuntu release through the SRU process. Whenever we need to make a change that modifies an existing machine type, we can make a "point-release" element in the Ubuntu names. This is not tied to a usual Ubuntu LTS point release, but to
```
s-makin: "LTSes" is not a valid word - we should say "LTS releases".
Should we link to the live migration page here?

s-makin:
```suggestion
These distro-specific machine types allow Ubuntu to support users who have deployed VMs on Ubuntu LTSes to be able to live migrate their VMs to the next LTS, dealing with whatever changes may have been introduced between one LTS and the next. Whenever we need to make a change that modifies an existing machine type, we can make a "point-release" element in the Ubuntu names. This is not tied to the usual Ubuntu LTS point releases, but to
```

s-makin: A bit confusingly worded overall

docs/explanation/virtualisation/qemu-machine-types.md
```diff
@@ -0,0 +1,168 @@
...
+anything introducing a delta to the machine type / `vmstate`. The format generally follows:
+
+```text
+$MACHINE_TYPE-$DISTRO-RELEASE[-v*]
+```
+
+This is an example of "point-release" machine types that have been added to address the [arch-capabilities bug](https://bugs.launchpad.net/bugs/2131822) in Noble:
```
s-makin:
```suggestion
This is an example of "point-release" machine types that have been added to address the {lpbug}`arch-capabilities bug <2131822>` in Noble:
```

docs/.custom_wordlist.txt
```diff
@@ -148,6 +153,7 @@ lookaside
 lookup
 lookups
 loopback
+LTSes
```
s-makin: Lets remove this, this isn't a valid word that should be excepted (see above)

docs/.custom_wordlist.txt
```diff
@@ -162,6 +168,7 @@ microcontroller
 Microvm
 microvm
 microvms
+migratability
```
s-makin: This is not a valid word, and should be removed

docs/redirects.txt
s-makin: Yes, they need to be kept forever once they're on the list, having both (as you've done here) is the correct method - thanks!

s-makin (changes requested):
Thanks for all your work on this. See suggestions/questions/comments inline :)

docs/explanation/virtualisation/live-migration.md
hector-cao: maybe "this is the second ..." is better ?

docs/explanation/virtualisation/qemu-machine-types.md
hector-cao: your proposal changes (the part "may have been introduced between one LTS and the next") the meaning of the sentence and I think we want to keep the original meaning;

docs/.custom_wordlist.txt
hector-cao: I have these spelling errors when I remove this (and the migratability):

 explanation/virtualisation/live-migration.md
 68:195  error  The word 'migratability' seems  Canonical.000-US-spellcheck 
                to be misspelled.                                           


 explanation/virtualisation/qemu-machine-types.md
 41:99  error  The word 'LTSes' seems to be    Canonical.000-US-spellcheck 
               misspelled.

docs/explanation/virtualisation/live-migration.md
hector-cao: level higher in the expressiveness of the libvirt VM definition syntax

hector-cao:
> Thanks for all your work on this. See suggestions/questions/comments inline :)

Thanks @s-makin , thanks for the review, please take a look at my follow ups.
I don't know why I cannot reply to your question so I put the answer here:

> What does "over the wire" mean? This feels quite colloquial

"over the wire" is our geek way of saying "on the network", "on the communication channel"

docs/explanation/virtualisation/live-migration.md
s-makin: I'd prefer if we could be specific rather than vague, and actually state what mechanism we're referring to

docs/explanation/virtualisation/live-migration.md
s-makin: Could we make this a bit clearer? I would never have guessed that that's what that referred to 🙃

docs/explanation/virtualisation/qemu-machine-types.md
s-makin: The original was quite confusingly worded though, I'm happy to keep the original meaning if we can find a way to make that more clear

docs/.custom_wordlist.txt
s-makin: I would expect so, they're not valid words though so they don't belong on the exception list. Let's find alternative ways to phrase so that we can avoid using them completely.

s-makin (changes requested):
Thanks @hector-cao, and sorry for the delay.

>> What does "over the wire" mean? This feels quite colloquial
> "over the wire" is our geek way of saying "on the network", "on the communication channel"

Such phrases can be quite easily misunderstood by anyone not familiar with them. Can we replace it with something more precise? 

I've added comments to your questions below, but it seems like the majority of my comments were not seen (I guess you were using the conversation tab, where most of the comments get hidden by default). Please take a look :)

docs/.custom_wordlist.txt
hector-cao: done

hector-cao:
> Thanks @hector-cao, and sorry for the delay.
> 
> > > What does "over the wire" mean? This feels quite colloquial
> > > "over the wire" is our geek way of saying "on the network", "on the communication channel"
> 
> Such phrases can be quite easily misunderstood by anyone not familiar with them. Can we replace it with something more precise?
> 
> I've added comments to your questions below, but it seems like the majority of my comments were not seen (I guess you were using the conversation tab, where most of the comments get hidden by default). Please take a look :)

Hey @s-makin,

I addressed the remaining comments you have on the MRs, please take a look and let me know if it is better

docs/explanation/virtualisation/live-migration.md
hector-cao: If users want to preserve the CPU features, they would need to restrict the migration to a host that would allow so

docs/explanation/virtualisation/live-migration.md
hector-cao: I used "Copy it into" instead of "Drop it"

docs/explanation/virtualisation/qemu-machine-types.md
hector-cao: I am afraid that this command 

$ virt-xml --edit --boot refresh-machine-type=yes <domain>

will make the machine type to switch from Ubuntu machine type to upstream version type, for example:

pc-q35-noble -> pc-q35-11.0

s-makin (approved):
Thanks for all the changes, lgtm :)

PR #757, opened 2026-07-23
Fixes for issue #755 related to the Managing software tutorial

### LINKED ISSUE #755: [Feedback]: Managing-software-tutorial: Fix references to noble and improve sentence organization in a couple of sections
### Page URL

https://ubuntu.com/server/docs/tutorial/managing-software/

### Select an option

- [x] I found the information, but it was confusing
- [x] I encountered a technical issue (e.g., broken link, image not loading)

### Issue Description

I checked the Managing software tutorial and found the following issues that need fixing

- [Line 54](https://github.com/canonical/ubuntu-server-documentation/blob/b9d3b55199196d56c2ca58ab033a6361dccef4b6/docs/tutorial/managing-software.md?plain=1#L54) : Refers to "noble" and the 24.04 LTS release though the example output above it shows "resolute". Also, the sentence promises to describe something later but the later section mostly refers to "resolute".
- [Line 66-69](https://github.com/canonical/ubuntu-server-documentation/blob/b9d3b55199196d56c2ca58ab033a6361dccef4b6/docs/tutorial/managing-software.md?plain=1#L66-L69): refers to *noble* package names and versions
- [Line 1055](https://github.com/canonical/ubuntu-server-documentation/blob/b9d3b55199196d56c2ca58ab033a6361dccef4b6/docs/tutorial/managing-software.md?plain=1#L1055) "noble" is used once though the rest of the paragraph refers to "resolute".

Some other points that I think could be improved.
- [line 118](https://github.com/canonical/ubuntu-server-documentation/blob/b9d3b55199196d56c2ca58ab033a6361dccef4b6/docs/tutorial/managing-software.md?plain=1#L118) "pay attention to it" got my attention the wrong way. The phrase is not very specific and because "in a script" is in brackets, I wondered how I should pay attention while the command is being executed. I suggest to remove the brackets around "in a script" and write something more specific, like -- "review which packages will be removed."
- [line 176](https://github.com/canonical/ubuntu-server-documentation/blob/b9d3b55199196d56c2ca58ab033a6361dccef4b6/docs/tutorial/managing-software.md?plain=1#L176) The topic changes to `apt-get` and `apt-cache` quite suddenly. Fix by beginning the sentence with something like: *While we used `apt` in the `apt show` command,*
- [About dependencies section](https://ubuntu.com/server/docs/tutorial/managing-software/#about-dependencies) Maybe the parts in brackets can be edited a little. I would explain why packages to remove are not shown in the example rather than indicate what will be explained in the next section.
- [Types of dependencies section](https://ubuntu.com/server/docs/tutorial/managing-software/#types-of-dependencies) The output of `apt show apache2` is missing. The descriptions for **depends**, **recommends** and **suggests** can be illustrated in the `show apache2` output. Maybe some sentences could be reorganized while making this change.
- Lastly, sometimes descriptions come after output examples in code block and sometimes before. It would be easier to read if it's always one or other.

VikramJohn:
### Description

Addresses the problems described in issue #755 

---

### Related issue

If this pull request addresses an issue, please add the issue number. This will
[close the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) automatically when your pull request is merged.

Example:

- Fixes: #755 

---

### Commit message for squash merge

We typically squash commits when merging. You can specify the commit message that should be used in this case if you wish:

[(optional) category] Brief description of changes made, and why

---

### AI usage

If you used AI to assist you in the creation of this PR, please let us know:

- The model used: Claude Sonnet 5

- How you used it: Spelling, grammar and formatting check

---

### Checklist

- [ X ] I have read and followed the [Ubuntu Server contributing guide](https://ubuntu.com/server/docs/contributing/).
- [ X ] My pull request is linked to an existing issue (if applicable).
- [ X ] I have tested my changes, and they work as expected.
- [ X ] I have disclosed my usage of AI

---

---

### DIFF
--- a/docs/tutorial/managing-software.md
+++ b/docs/tutorial/managing-software.md
@@ -263,19 +263,6 @@ We can see, using `apt show`, exactly which packages fall into each of these cat
 :host:
 :dir:
 apt show apache2
-```
-
-```{terminal}
-:output-only:
-[...]
-Depends: apache2-bin (= 2.4.66-2ubuntu2.4), apache2-data (= 2.4.66-2ubuntu2.4), apache2-utils (= 2.4.66-2ubuntu2.4), media-types, procps, perl:any
-Recommends: ssl-cert
-```{terminal}
-:copy:
-:user:
-:host:
-:dir:
-apt show apache2
 
 [...]
 Depends: apache2-bin (= 2.4.66-2ubuntu2.4), apache2-data (= 2.4.66-2ubuntu2.4), apache2-utils (= 2.4.66-2ubuntu2.4), media-types, procps, perl:any
--- a/docs/tutorial/managing-software.md
+++ b/docs/tutorial/managing-software.md
@@ -340,10 +340,7 @@ But actually, if we run this command:
 :host:
 :dir:
 sudo apt install apache2 --install-suggests
-```
 
-```{terminal}
-:output-only:
 [...]
 Installing dependencies:
   apache2-bin              apache2-utils            libaprutil1-ldap
@@ -357,7 +354,7 @@ Summary:
   Space needed: 34.1 MB / 1328 MB available
 ```
 
-By comparing the amount of space needed with "suggests" included, to the previous output, we can see that there's a considerable increase! Sometimes, including suggested packages will cause the package to not be installed due to a lack of space on the system:
+By comparing the amount of space needed with "suggests" included, to the previous output, we can see that there's a considerable increase! Sometimes, including suggested packages will cause the package to not be installed due to a lack of space on the system.
 
 This is because each of these suggested packages also comes with their own lists of dependencies, including suggested packages, all of which would *also* be installed. It's perhaps clear to see why this is not the default setting!
 
@@ -706,7 +703,7 @@ dpkg --verify apache2
 You might have noticed there's a "c" next to the top line but not the bottom -- the "c" shows the file is a conffile.
 :::
 
-`dpkg` can tell that the file has been changed, but won't tell us what the change was. However, since the file in question is not a conffile, we know that the change *won't be preserved* if we upgrade the package. This means that we can overwrite the changes and restore the default package content by "reinstalling" Apache2. We can use the `--reinstall` flag to force `apt` to re-unpack all of the default content.
+`dpkg` can tell that the file has been changed, but won't tell us what the change was. However, since the file in question is not a conffile, we know that the change *won't be preserved* if we upgrade the package. This means that we can overwrite the changes and restore the default package content by "reinstalling" Apache2. We can use the `--reinstall` flag to force `apt` to re-unpack all of the default content:
 
 ```{terminal}
 :copy:
@@ -973,7 +970,7 @@ Error: Unable to satisfy dependencies. Reached two conflicting assignments:
       - apache2-bin:amd64=2.4.66-2ubuntu2 is not selected for install
 ```
 
-So, all we need to do is first install the dependencies, and then run the install command again. Remember that we can install multiple packages at once by separating them with spaces. We can also break the command over multiple lines using backslashes (`\`) to make it easier to read, but it will still be run as a single command.
+So, all we need to do is first install the dependencies, and then run the install command again. Remember that we can install multiple packages at once by separating them with spaces. We can also break the command over multiple lines using backslashes (`\`) to make it easier to read, but it will still be run as a single command:
 
 ```{terminal}
 :copy:
--- a/docs/tutorial/managing-software.md
+++ b/docs/tutorial/managing-software.md
@@ -270,6 +270,16 @@ apt show apache2
 [...]
 Depends: apache2-bin (= 2.4.66-2ubuntu2.4), apache2-data (= 2.4.66-2ubuntu2.4), apache2-utils (= 2.4.66-2ubuntu2.4), media-types, procps, perl:any
 Recommends: ssl-cert
+```{terminal}
+:copy:
+:user:
+:host:
+:dir:
+apt show apache2
+
+[...]
+Depends: apache2-bin (= 2.4.66-2ubuntu2.4), apache2-data (= 2.4.66-2ubuntu2.4), apache2-utils (= 2.4.66-2ubuntu2.4), media-types, procps, perl:any
+Recommends: ssl-cert
 Suggests: apache2-doc, apache2-suexec-pristine | apache2-suexec-custom, ufw, www-browser
 [...]
 ```
--- a/docs/tutorial/managing-software.md
+++ b/docs/tutorial/managing-software.md
@@ -51,7 +51,7 @@ Hit:4 http://security.ubuntu.com/ubuntu resolute-security InRelease
 19 packages can be upgraded. Run 'apt list --upgradable' to see them.
 ```
 
-As we can see, it checks ("hits") the various archives (**pockets**) that updates can come from for the 24.04 LTS release (`noble-security`, `noble`, `noble-updates` and `noble-backports` -- remember these, as we'll come back to them later). It has found some packages that can be upgraded to newer versions. If we want to see which packages those are, we can run the command hinted in the output:
+As we can see, it checks ("hits") the various archives (**pockets**) that updates can come from for the 26.04 LTS release (`resolute`, `resolute-updates`, `resolute-backports`, and `resolute-security` -- remember these, as we'll come back to them later). It has found some packages that can be upgraded to newer versions. If we want to see which packages those are, we can run the command hinted in the output:
 
 ```{terminal}
 :copy:
@@ -63,10 +63,10 @@ apt list --upgradable
 
 The output tells us:
 
-- the package name and where the update will come from (e.g. `base-files/noble-updates`),
-- the most up-to-date package version available (e.g. `13ubuntu10.1`)
+- the package name and where the update will come from (e.g. `bind9-dnsutils` package from `resolute-updates` and `resolute-security`),
+- the most up-to-date package version available (e.g. `1:9.20.18-1ubuntu2.1`),
 - the hardware version the update is for (e.g. `amd64`), and
-- what package version is currently installed (e.g. `13ubuntu10`)
+- what package version is currently installed (e.g. `1:9.20.18-1ubuntu2`)
 
 The specific packages included in this list changes over time, so the exact packages shown will be different, but the output will be structured like this:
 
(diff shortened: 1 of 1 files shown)


docs/tutorial/managing-software.md
```diff
@@ -109,13 +109,13 @@ In the output, we'll see where `apt upgrade` is fetching the upgrade from for ea
...
 :::
 
-It's important to know that `apt upgrade` will only handle packages that can be straightforwardly upgraded. If the package has **dependency** issues (i.e., the version you have "depends" on other packages that also need to be added, upgraded or removed), you would need to use `sudo apt dist-upgrade` instead. The `dist-upgrade` command is able to resolve conflicts between package versions, but it *could* end up removing some packages -- so although `apt upgrade` is safe to use unattended (in a script, for example), you should only use `dist-upgrade` when you can pay attention to it.
+It's important to know that `apt upgrade` will only handle packages that can be straightforwardly upgraded. If the package has **dependency** issues (i.e., the version you have "depends" on other packages that also need to be added, upgraded or removed), you would need to use `sudo apt dist-upgrade` instead. The `dist-upgrade` command is able to resolve conflicts between package versions, but it *could* end up removing some packages. So it's safe to use `-y` (or a similar flag) with `apt upgrade`, but you should only run `dist-upgrade` when you can review which packages may be removed.
```
s-makin: This has changed the meaning - we were intentionally explicit about what command is safe to run unattended vs when you must be there to watch what's happening.

What in particular was confusing about the original wording that led you to change it? We can find a wording that addresses the problem you found with it that doesn't change the intent of the warning.

docs/tutorial/managing-software.md
```diff
@@ -173,7 +173,7 @@ Description: parameter calculator for IPv4 addresses
  webserver.
 ```
 
-In many places, you will see reference to `apt-get` and `apt-cache` instead of `apt`. Historically, the *database* part of APT was accessed using `apt-cache` (e.g. `apt-cache show ipcalc`), and the *packages* part of APT used `apt-get` (e.g. `apt-get install ipcalc`).
+While we used `apt` in the `apt show` command, in many places you will see reference to `apt-get` and `apt-cache` instead of `apt`. Historically, the *database* part of APT was accessed using `apt-cache` (e.g. `apt-cache show ipcalc`), and the *packages* part of APT used `apt-get` (e.g. `apt-get install ipcalc`).
```
s-makin, line 176: I'm not sure this is needed - it's the sort of thing I would edit out if I saw it in a draft, because it doesn't add much to the clarity of the sentence (just more words). I'm happy to be convinced otherwise if you feel very strongly that it's needed :)

docs/tutorial/managing-software.md
```diff
@@ -263,7 +265,16 @@ We can see, using `apt show`, exactly which packages fall into each of these cat
 apt show apache2
 ```
 
-If we look only at the sections on dependencies, we can see that `ssl-cert` is a recommended package:
+```{terminal}
+:output-only:
+[...]
+Depends: apache2-bin (= 2.4.66-2ubuntu2.4), apache2-data (= 2.4.66-2ubuntu2.4), apache2-utils (= 2.4.66-2ubuntu2.4), media-types, procps, perl:any
+Recommends: ssl-cert
+Suggests: apache2-doc, apache2-suexec-pristine | apache2-suexec-custom, ufw, www-browser
```
s-makin:
```suggestion
```{terminal}
:copy:
:user:
:host:
:dir:
apt show apache2

[...]
Depends: apache2-bin (= 2.4.66-2ubuntu2.4), apache2-data (= 2.4.66-2ubuntu2.4), apache2-utils (= 2.4.66-2ubuntu2.4), media-types, procps, perl:any
Recommends: ssl-cert
Suggests: apache2-doc, apache2-suexec-pristine | apache2-suexec-custom, ufw, www-browser
```

s-makin: If we're taking out the text in the middle, we can join the two terminal blocks :)

docs/tutorial/managing-software.md
```diff
@@ -323,8 +332,6 @@ But actually, if we run this command:
 sudo apt install apache2 --install-suggests
 ```
```
s-makin: Same as above, if there's no text in the middle to separate the command and output, we can just show them in one terminal block

docs/tutorial/managing-software.md
```diff
@@ -340,6 +347,8 @@ Summary:
   Space needed: 34.1 MB / 1328 MB available
 ```
 
+By comparing the amount of space needed with "suggests" included, to the previous output, we can see that there's a considerable increase! Sometimes, including suggested packages will cause the package to not be installed due to a lack of space on the system:
```
s-makin:
```suggestion
By comparing the amount of space needed with "suggests" included, to the previous output, we can see that there's a considerable increase! Sometimes, including suggested packages will cause the package to not be installed due to a lack of space on the system.
```

docs/tutorial/managing-software.md
```diff
@@ -687,7 +696,7 @@ dpkg --verify apache2
 You might have noticed there's a "c" next to the top line but not the bottom -- the "c" shows the file is a conffile.
 :::
 
-`dpkg` can tell that the file has been changed, but won't tell us what the change was. However, since the file in question is not a conffile, we know that the change *won't be preserved* if we upgrade the package. This means that we can overwrite the changes and restore the default package content by "reinstalling" Apache2:
+`dpkg` can tell that the file has been changed, but won't tell us what the change was. However, since the file in question is not a conffile, we know that the change *won't be preserved* if we upgrade the package. This means that we can overwrite the changes and restore the default package content by "reinstalling" Apache2. We can use the `--reinstall` flag to force `apt` to re-unpack all of the default content.
```
s-makin:
```suggestion
`dpkg` can tell that the file has been changed, but won't tell us what the change was. However, since the file in question is not a conffile, we know that the change *won't be preserved* if we upgrade the package. This means that we can overwrite the changes and restore the default package content by "reinstalling" Apache2. We can use the `--reinstall` flag to force `apt` to re-unpack all of the default content:
```

docs/tutorial/managing-software.md
```diff
@@ -954,7 +963,7 @@ Error: Unable to satisfy dependencies. Reached two conflicting assignments:
       - apache2-bin:amd64=2.4.66-2ubuntu2 is not selected for install
 ```
 
-So, all we need to do is first install the dependencies, and then run the install command again. Remember that we can install multiple packages at once by separating them with spaces:
+So, all we need to do is first install the dependencies, and then run the install command again. Remember that we can install multiple packages at once by separating them with spaces. We can also break the command over multiple lines using backslashes (`\`) to make it easier to read, but it will still be run as a single command.
```
s-makin:
```suggestion
So, all we need to do is first install the dependencies, and then run the install command again. Remember that we can install multiple packages at once by separating them with spaces. We can also break the command over multiple lines using backslashes (`\`) to make it easier to read, but it will still be run as a single command:
```

s-makin (changes requested):
Overall looks pretty good, thanks for making these changes! I've left a few suggestions and questions inline for your consideration.

docs/tutorial/managing-software.md
VikramJohn: Well, maybe it's because I didn't approach the tutorial with the notion that all commands that make changes need confirmation before they are executed.:) So when I first read it, I was like "pay attention" ... how -- should I watch the execution and try to intervene? I think with the earlier version the specific point at which the command is executed is lost in "dist-upgrade could end up removing some packages if you don't pay attention to it".

Once I understood that the "paying attention" is about confirming / not confirming execution, I tried to put that understanding into the text and be more specific about when or how each command is ok to use.

Did I make it too specific? Does it now exclude some possible usages?

docs/tutorial/managing-software.md
VikramJohn: It felt like a divergence in the narrative because the context at that point is not `apt` itself, but `apt show`. I think if a user isn't expecting to see apt-get and apt-cache discussed (maybe that's just me), this feels like a strange break. Maybe we could just add a H4 heading such as "apt-get and apt-cache" or "alternatives to apt" to section that part of instead of adding the "while" clause.

docs/tutorial/managing-software.md
VikramJohn: Fixed

docs/tutorial/managing-software.md
VikramJohn: Fixed

docs/tutorial/managing-software.md
VikramJohn: Fixed

docs/tutorial/managing-software.md
VikramJohn: Done

docs/tutorial/managing-software.md
VikramJohn: Joined them

docs/tutorial/managing-software.md
s-makin: I'd prefer not to add any extra headings for such a small section, otherwise it would unbalance the flow on the page. On reflection, I think I'd rather accept your change as-is. I can see what you mean about the sudden shift now that you mention it, and I think you were right in how you handled it.

docs/tutorial/managing-software.md
s-makin: The problem with the new version is that it implies that using `-y` is what's being discussed (it's not). The difference we wanted to illustrate is when you would or would not use the two different commands. `apt upgrade` is the more defensive one, which makes it safe to run without watching it (because it will only touch straightfoward upgrades, anything that might have conflicts gets left alone). `dist-upgrade`, by contrast, can handle more complicated cases but it might have undesired behaviour so you have to watch it to catch those cases (i.e., you shouldn't let it run unattended).

s-makin (commented):
Thank you for resolving the previous issues, I've left a reply on the final question left outstanding.

docs/tutorial/managing-software.md
VikramJohn: Yes, I see the problem with it. Also, this version indirectly suggests "-y" or a similar flag is universal, which is probably confusing and untrue. I've modified the line to something much closer to the original. Basically I've replaced "pay attention to it" with "review which packages may be removed", which I feel is a bit more concrete (and which I see is exactly what I proposed while creating the issue :))

docs/tutorial/managing-software.md
```diff
@@ -109,13 +109,13 @@ In the output, we'll see where `apt upgrade` is fetching the upgrade from for ea
...
 :::
 
-It's important to know that `apt upgrade` will only handle packages that can be straightforwardly upgraded. If the package has **dependency** issues (i.e., the version you have "depends" on other packages that also need to be added, upgraded or removed), you would need to use `sudo apt dist-upgrade` instead. The `dist-upgrade` command is able to resolve conflicts between package versions, but it *could* end up removing some packages -- so although `apt upgrade` is safe to use unattended (in a script, for example), you should only use `dist-upgrade` when you can pay attention to it.
+It's important to know that `apt upgrade` will only handle packages that can be straightforwardly upgraded. If the package has **dependency** issues (i.e., the version you have "depends" on other packages that also need to be added, upgraded or removed), you would need to use `sudo apt dist-upgrade` instead. The `dist-upgrade` command is able to resolve conflicts between package versions, but it *could* end up removing some packages. So while `apt upgrade` is safe to use unattended (in a script, for example), you should only use `dist-upgrade` when you can review which packages may be removed.
```
s-makin, line 118: Thanks, this is much better - clear, but still keeps the original intention

s-makin (approved):
Thanks for all the changes @VikramJohn, overall this is a really great contribution and I appreciate your patience with the various rounds of review (especially the last one as I've been out of the office). Thank you for contributing to the Server docs, very happy to merge this! :)

VikramJohn:
@s-makin 
Well, I was grateful for the review because I wasn't sure had the full picture while suggesting edits. The wait was no trouble; I guessed that you were out of office. 
Thank you! I really enjoyed working on this tutorial. In general, I like exploring content in this way, both using the information and evaluating it from an author's perspective. 
I look forward to contributing more.

PR #775, opened 2026-08-18
reference/other-tools: include rust-coreutils in sudo-rs

### LINKED ISSUE #774: [Feedback]: Other tools page for rust-coreutils
### Page URL

https://ubuntu.com/server/docs/reference/other-tools/

### Select an option

- [x] I couldn't find what I was looking for

### Issue Description

Since 26.10 will move to use 100% `rust-coreutils` by default, it may be a good idea to explain possible divergences and how users can revert back to GNU.

BAMF0:
### Description

This pull request adds necessary context for `rust-coreutils` in the Ubuntu Server docs, including how to switch to GNU and which tools are still using GNU in specific releases.

It provides a reference point for users coming from previous LTS versions where GNU was the coreutils provider.

---

### Related issue

- Fixes: #774 

---

### Checklist

- [X] I have disclosed my usage of AI

---

WIP: this PR intentionally does not restructure or rename the current `sudo-rs.md` file. The appropriate formatting is to be discussed.

### DIFF
--- a/docs/reference/other-tools/sudo-rs.md
+++ b/docs/reference/other-tools/sudo-rs.md
@@ -1,13 +1,58 @@
 ---
 myst:
   html_meta:
-    description: "Reference documentation for sudo-rs outlining differences to the former sudo.ws, to guide users of Ubuntu Server on the new implementation."
+    description: "Reference documentation for migrated tools outlining differences to the former, to guide users of Ubuntu Server on the new implementations."
 ---
 
-(sudo-rs)=
-# sudo-rs
+# Migrated tools
 
-This page serves as a reference for the key differences between `sudo.ws` and `sudo-rs`.
+The “oxidisation” of Ubuntu (see [Ubuntu Discourse](https://discourse.ubuntu.com/t/carefully-but-purposefully-oxidising-ubuntu/56995)) has changed the providers of a number of packages that are installed by default.
+This page provides guidance for Ubuntu Server users on the new implementations
+
+(rust-coreutils)=
+## rust-coreutils
+
+uutils is the default coreutils provider as of Ubuntu 25.10. This section lists
+helpful resources and tips regarding the migration.
+
+Refer to the [uutils Coreutils Documentation](https://uutils.org/coreutils/docs/index.html) for information about uutils.
+
+### Exceptions per releases
+
+Due to known incompatibilities, some utilities in `rust-coreutils` are still
+provided by GNU for the Ubuntu 25.10 and 26.04 releases. Redirected utilities
+per release are as follows:
+
+| Release | Utilities still provided by GNU |
+|---|---|
+| **Ubuntu 25.10** ([reference][ubuntu-25.10-reference]) | `chmod`, `chown`, `cp`, `df`, `mv`, `rm`, `true` |
+| **Ubuntu 26.04** ([reference][ubuntu-26.04-reference]) | `cp`, `df`, `mv`, `rm`, `true` |
+| **Ubuntu 26.10 and later** | — |
+
+[ubuntu-25.10-reference]: https://git.launchpad.net/ubuntu/+source/coreutils-from/tree/debian/coreutils-from-uutils.links?h=ubuntu/questing-devel
+[ubuntu-26.04-reference]: https://git.launchpad.net/ubuntu/+source/coreutils-from/tree/debian/coreutils-from-uutils.links?h=ubuntu/resolute-devel
+
+### Changing provider
+
+The installed `coreutils-from-*` package determines which coreutils provider is
+used on the system.
+
+**To switch to GNU coreutils:**
+
+```shell
+sudo apt install coreutils-from-gnu coreutils-from-uutils- --allow-remove-essential
+```
+
+**To switch to back to rust-coreutils:**
+
+```shell
+sudo apt install coreutils-from-uutils coreutils-from-gnu- --allow-remove-essential
+```
+
+(sudo-rs)=
+## sudo-rs
+
+This section serves as a reference for the key differences between `sudo.ws` and `sudo-rs`.
 
 Note: Both projects are under active development, so it is not possible to maintain a fully up-to-date list of differences.
 This is a list of major differences as of the Ubuntu 25.10 and 26.04 releases, to help users upgrading to those.
@@ -16,7 +61,7 @@ For the most accurate and current information, refer to
 `sudo-rs --help` for a list of supported options in your installed version.
 Refer to `man sudoers-rs` for the `/etc/sudoers` configuration options supported by `sudo-rs`.
 
-## Differences
+### Differences
 
 1. Start with the official documentation from the `sudo-rs` project.
 
--- a/docs/reference/other-tools/sudo-rs.md
+++ b/docs/reference/other-tools/sudo-rs.md
@@ -6,12 +6,14 @@ myst:
 
 # System utility replacements
 
-Starting with Ubuntu 25.10, the "oxidization" of Ubuntu (see [Ubuntu Discourse][oxiziding-ubuntu-reference]) changed the providers of the system utilities
-`coreutils` and `sudo` to new Rust-based implementations. Previous Ubuntu
-releases retain their existing default providers.
+Starting with Ubuntu 25.10, the "oxidization" of Ubuntu (see [Ubuntu
+Discourse](https://discourse.ubuntu.com/t/carefully-but-purposefully-oxidising-ubuntu/56995))
+changed the providers of the system utilities `coreutils` and `sudo` to new
+Rust-based implementations. Previous Ubuntu releases retain their existing
+default providers.
 
-- **`coreutils`**: changed from [GNU Coreutils][gnu-coreutils-manual-reference] to [uutils Coreutils][uutils-coreutils-manual-reference]
-- **`sudo`**: changed from [`sudo.ws`][sudo-ws-manual-reference] to [`sudo-rs`][sudo-rs-project-reference]
+- **`coreutils`**: changed from [GNU Coreutils](https://www.gnu.org/software/coreutils/manual) to [uutils Coreutils](https://uutils.org/coreutils/docs)
+- **`sudo`**: changed from [`sudo.ws`](https://www.sudo.ws/docs/man/sudo.man/) to [`sudo-rs`](https://github.com/trifectatechfoundation/sudo-rs)
 
 For most users, no changes are required for normal usage. The new Rust-based
 implementations are intended to be drop-in replacements for the existing
@@ -24,12 +26,6 @@ where they are the default providers. In newer releases where the Rust-based
 implementations are the default, **GNU Coreutils and sudo.ws remain available
 as providers**, while the Rust-based implementations are the primary ones
 going forward.
-
-[oxiziding-ubuntu-reference]: https://discourse.ubuntu.com/t/carefully-but-purposefully-oxidising-ubuntu/56995
-[uutils-coreutils-manual-reference]: https://uutils.org/coreutils/docs
-[gnu-coreutils-manual-reference]: https://www.gnu.org/software/coreutils/manual
-[sudo-ws-manual-reference]: https://www.sudo.ws/docs/man/sudo.man/
-[sudo-rs-project-reference]: https://github.com/trifectatechfoundation/sudo-rs
 
 (rust-coreutils)=
 ## rust-coreutils
@@ -86,13 +82,21 @@ providers.
 
 **To switch to GNU Coreutils:**
 
-```shell
+```{terminal}
+:copy:
+:user:
+:host:
+:dir:
 sudo apt install coreutils-from-gnu coreutils-from-uutils- --allow-remove-essential
 ```
 
 **To switch back to `rust-coreutils`:**
 
-```shell
+```{terminal}
+:copy:
+:user:
+:host:
+:dir:
 sudo apt install coreutils-from-uutils coreutils-from-gnu- --allow-remove-essential
 ```
 
--- a/docs/reference/other-tools/sudo-rs.md
+++ b/docs/reference/other-tools/sudo-rs.md
@@ -8,9 +8,10 @@ myst:
 
 Starting with Ubuntu 25.10, the "oxidization" of Ubuntu (see [Ubuntu Discourse][oxiziding-ubuntu-reference]) changed the providers of the system utilities
 `coreutils` and `sudo` to new Rust-based implementations. Previous Ubuntu
-releases are unaffected.
-- `coreutils`: changed from [GNU Coreutils][gnu-coreutils-manual-reference] to [uutils Coreutils][uutils-coreutils-manual-reference]
-- `sudo`: changed from [`sudo.ws`][sudo-ws-manual-reference] to [`sudo-rs`][sudo-rs-project-reference]
+releases retain their existing default providers.
+
+- **`coreutils`**: changed from [GNU Coreutils][gnu-coreutils-manual-reference] to [uutils Coreutils][uutils-coreutils-manual-reference]
+- **`sudo`**: changed from [`sudo.ws`][sudo-ws-manual-reference] to [`sudo-rs`][sudo-rs-project-reference]
 
 For most users, no changes are required for normal usage. The new Rust-based
 implementations are intended to be drop-in replacements for the existing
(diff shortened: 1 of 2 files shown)


docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,58 @@
 ---
 myst:
   html_meta:
-    description: "Reference documentation for sudo-rs outlining differences to the former sudo.ws, to guide users of Ubuntu Server on the new implementation."
+    description: "Reference documentation for migrated tools outlining differences to the former, to guide users of Ubuntu Server on the new implementations."
```
s-makin: "migrated tools" is a bit vague. Is there a more specific/descriptive phrasing we can use to guide users to understand specifically what tools have been migrated?

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,58 @@
...
 ---
 
+# Migrated tools
+
+The “oxidisation” of Ubuntu (see [Ubuntu Discourse](https://discourse.ubuntu.com/t/carefully-but-purposefully-oxidising-ubuntu/56995)) has changed the providers of a number of packages that are installed by default.
```
s-makin:
```suggestion
The [“oxidization” of Ubuntu](https://discourse.ubuntu.com/t/carefully-but-purposefully-oxidising-ubuntu/56995) has changed the providers of a number of packages that are installed by default.
```

s-makin: It would be good to be more specific here, as well. Instead of "a number of packages", let's state explicitly which packages have changed, and from which provider to what. In the intro section, we need to help a reader to understand if this page is relevant to their needs or not.

I'll leave it up to you how you want to describe the changes, it doesn't need to be super detailed for the intro but it should be specific enough to help a user who doesn't follow Ubuntu news to understand that this is about certain system utilities, and what the migration means for them.

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,58 @@
...
+This page provides guidance for Ubuntu Server users on the new implementations
```
s-makin:
```suggestion
This page provides guidance on the new implementations.
```

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,58 @@
...
+
+(rust-coreutils)=
+## rust-coreutils
+
+uutils is the default coreutils provider as of Ubuntu 25.10. This section lists
```
s-makin: This is a bit confusing. The title is `rust-coreutils`, but then you're talking about `uutils` and coreutils, but later switch back again. I think we need a little more detail to make this switch clearer:
- What was the default provider before 25.10?
- What's the difference between rust-coreutils and uutils?
- Are there any Ubuntu releases where uutils is not available? (i.e. has it been backported)
- Will the previous default be deprecated at some point or is it going to remain available?
- What are the implications of the change/does the user need to do anything different?
(some of these questions might just be for me, but I'm approaching this from the PoV of someone who isn't familiar with this)

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,58 @@
...
+helpful resources and tips regarding the migration.
+
+Refer to the [uutils Coreutils Documentation](https://uutils.org/coreutils/docs/index.html) for information about uutils.
```
s-makin: The spellchecker isn't happy about uutils or coreutils where they appear in the text. Can we put these into backticks to render them as inline code? (this avoids triggering the spellchecker)

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,58 @@
...
+
+### Exceptions per releases
+
+Due to known incompatibilities, some utilities in `rust-coreutils` are still
+provided by GNU for the Ubuntu 25.10 and 26.04 releases. Redirected utilities
```
s-makin:
```suggestion
provided by GNU for the Ubuntu 25.10 and Ubuntu 26.04 LTS releases. Redirected utilities
```

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,58 @@
...
+per release are as follows:
+
+| Release | Utilities still provided by GNU |
+|---|---|
+| **Ubuntu 25.10** ([reference][ubuntu-25.10-reference]) | `chmod`, `chown`, `cp`, `df`, `mv`, `rm`, `true` |
+| **Ubuntu 26.04** ([reference][ubuntu-26.04-reference]) | `cp`, `df`, `mv`, `rm`, `true` |
```
s-makin:
```suggestion
| **Ubuntu 26.04 LTS** ([reference][ubuntu-26.04-reference]) | `cp`, `df`, `mv`, `rm`, `true` |
```

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,58 @@
...
+| **Ubuntu 26.10 and later** | — |
+
+[ubuntu-25.10-reference]: https://git.launchpad.net/ubuntu/+source/coreutils-from/tree/debian/coreutils-from-uutils.links?h=ubuntu/questing-devel
+[ubuntu-26.04-reference]: https://git.launchpad.net/ubuntu/+source/coreutils-from/tree/debian/coreutils-from-uutils.links?h=ubuntu/resolute-devel
+
+### Changing provider
+
+The installed `coreutils-from-*` package determines which coreutils provider is
+used on the system.
```
s-makin: If there are any restrictions per release (i.e. any where the package is not available) please mention them here to make it clear before any user tries to use the commands on an incompatible release

s-makin (changes requested):
Nice draft - I've left some suggestions, questions and comments inline :)

docs/reference/other-tools/sudo-rs.md
BAMF0: Hmm good question, I agree. One idea is "oxidised" tools, but that also feels a bit vague. Let me think about this.

docs/reference/other-tools/sudo-rs.md
BAMF0: True! This phrasing was used in order to also apply if more oxidized packages where added, but I see now that this is at the cost of clarity.

docs/reference/other-tools/sudo-rs.md
BAMF0: Good points! Regarding `rust-coreutils` vs. `uutils`: I was a bit unsure how to put it while still remaining correct. What makes it a bit confusing is that "uutils coreutils" is the *actual* implementation, and `rust-coreutils` is the packaged name for it we use. I.e. `uutils` is a collection of CLI utilities rewritten in Rust. 

We use the package `coreutils-from-uutils` for setting the `rust-coreutils` as default. I preferred `rust-coreutils` as the title since that is what most people will think of it as/relate to.

One solution could be to move the mentioning of `uutils` and simply state the first sentence as something like "A Rust-based implementation of the coreutils" instead.

docs/reference/other-tools/sudo-rs.md
BAMF0: Sure thing!

docs/reference/other-tools/sudo-rs.md
BAMF0: Ah yes, thanks! Will do

docs/reference/other-tools/sudo-rs.md
BAMF0: I tested it out and I fear it has an effect on the readability of the text. Is it possible to add these to the vocabulary or add some ignore for them?

s-makin: the simplest way is probably to add them to the `docs/.custom_wordlist.txt` file, otherwise it makes the source file unreadable too :D

docs/reference/other-tools/sudo-rs.md
BAMF0: Suggestion in: https://github.com/canonical/ubuntu-server-documentation/pull/775/commits/f4fab366f3e0bc1cf3aa980afd3d849fcd892519

docs/reference/other-tools/sudo-rs.md
BAMF0: Fixed in: https://github.com/canonical/ubuntu-server-documentation/pull/775/commits/f4fab366f3e0bc1cf3aa980afd3d849fcd892519

docs/reference/other-tools/sudo-rs.md
BAMF0: Fixed in: https://github.com/canonical/ubuntu-server-documentation/pull/775/commits/f4fab366f3e0bc1cf3aa980afd3d849fcd892519

docs/reference/other-tools/sudo-rs.md
BAMF0: Mostly fixed in: https://github.com/canonical/ubuntu-server-documentation/pull/775/commits/f4fab366f3e0bc1cf3aa980afd3d849fcd892519

Still have to answer:

> Are there any Ubuntu releases where uutils is not available? (i.e. has it been backported)

> Will the previous default be deprecated at some point or is it going to remain available?

> What are the implications of the change/does the user need to do anything different?

docs/reference/other-tools/sudo-rs.md
BAMF0: Fixed in: https://github.com/canonical/ubuntu-server-documentation/commit/f4fab366f3e0bc1cf3aa980afd3d849fcd892519

docs/reference/other-tools/sudo-rs.md
BAMF0: Fixed in: https://github.com/canonical/ubuntu-server-documentation/commit/f4fab366f3e0bc1cf3aa980afd3d849fcd892519

docs/reference/other-tools/sudo-rs.md
BAMF0: Fixed in: https://github.com/canonical/ubuntu-server-documentation/commit/f4fab366f3e0bc1cf3aa980afd3d849fcd892519

docs/reference/other-tools/sudo-rs.md
BAMF0: Hahah exactly! I looked it up and there would have to be horrible Vale HTML for each word we want to ignore.

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,86 @@
...
+| **Ubuntu 24.04 LTS and earlier** | **All** |
+| **Ubuntu 25.10** | `chmod`, `chown`, `cp`, `df`, `mv`, `rm`, `true` |
+| **Ubuntu 26.04 LTS** | `cp`, `df`, `mv`, `rm`, `true` |
+| **Ubuntu 26.10 and later** | **None** |
+
+:::{note}
+Although the `rust-coreutils` package is available in Ubuntu 24.04 LTS, it
+cannot be configured as the default `coreutils`. The
+`coreutils-from-<provider>` packages required to switch the provider are only
+available starting with Ubuntu 26.04 LTS.
```
BAMF0: Should this be 25.10? I'm a bit uncertain since that's EoL

docs/reference/other-tools/sudo-rs.md
BAMF0: Fixed now in: https://github.com/canonical/ubuntu-server-documentation/pull/775/commits/13ebccc1c8a9a1a8ef02eee06165c252841d549a

docs/reference/other-tools/sudo-rs.md
BAMF0: Fixed in: https://github.com/canonical/ubuntu-server-documentation/pull/775/commits/168756dabacdb608e39e3a07cc8a22e60defebf2

BAMF0:
There we go! I have made some changes based on your feedback @s-makin and I must say, the feedback was very valuable indeed.

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,126 @@
 ---
 myst:
   html_meta:
-    description: "Reference documentation for sudo-rs outlining differences to the former sudo.ws, to guide users of Ubuntu Server on the new implementation."
+    description: "Reference documentation for Ubuntu Server’s Rust-based replacements for sudo and GNU coreutils, highlighting differences, release-specific exceptions, and important changes for users"
```
s-makin: This is now a bit too long - we need to cut this down to a max of 160 characters :/

s-makin:
```suggestion
    description: "Reference page for Ubuntu’s Rust-based replacements for sudo and GNU coreutils, highlighting differences, release-specific exceptions, and important changes."
```

s-makin: This cuts it down to 156 - lmk if this works for you

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,126 @@
...
+as before. This page describes known differences, incompatibilities, and other
+considerations that may affect specific use cases.
+
+GNU Coreutils and `sudo.ws` continue to receive maintenance in Ubuntu releases
+where they are the default providers. In newer releases where the Rust-based
+implementations are the default, **GNU Coreutils and sudo.ws remain available
+as providers**, while the Rust-based implementations are the primary ones
+going forward.
+
+[oxiziding-ubuntu-reference]: https://discourse.ubuntu.com/t/carefully-but-purposefully-oxidising-ubuntu/56995
```
s-makin: We don't use this convention of link gathering in the Server docs - links should be at the place they appear in the text, otherwise it can become difficult to find them, with no real gain in source legibility. Please redistribute these links to where they're found in the doc.

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,126 @@
 ---
 myst:
   html_meta:
-    description: "Reference documentation for sudo-rs outlining differences to the former sudo.ws, to guide users of Ubuntu Server on the new implementation."
+    description: "Reference documentation for Ubuntu Server’s Rust-based replacements for sudo and GNU coreutils, highlighting differences, release-specific exceptions, and important changes for users"
 ---
 
+# System utility replacements
+
+Starting with Ubuntu 25.10, the "oxidization" of Ubuntu (see [Ubuntu Discourse][oxiziding-ubuntu-reference]) changed the providers of the system utilities
```
s-makin:
```suggestion
Starting with Ubuntu 25.10, the ["oxidization" of Ubuntu](https://discourse.ubuntu.com/t/carefully-but-purposefully-oxidising-ubuntu/56995) changed the providers of the system utilities
```

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,126 @@
...
+:::{note}
+`update-alternatives` is not currently suitable for switching between GNU
+Coreutils and uutils Coreutils, as alternatives are not safe for `Essential`
+packages. Instead, use the `coreutils-from-<provider>` packages to switch
+providers.
+:::
+
+**To switch to GNU Coreutils:**
+
+```shell
```
s-makin: Please use terminal blocks instead of markdown code blocks. I can't paste an example here because GitHub doesn't render them properly in this preview, but you can see [examples in this page](https://github.com/canonical/ubuntu-server-documentation/blob/main/docs/tutorial/welcome-to-the-terminal.md) that you can copy/paste

s-makin (changes requested):
Thanks for the first round of changes, I've left some additional comments and suggestions inline.

As a side note, please ignore the spellcheck failures, there seems to have been a recent change in the default behaviour so it's now picking up terms in code blocks (it shouldn't be). None of the failures apply to this PR.

docs/reference/other-tools/sudo-rs.md
BAMF0: Thanks! I think that works well.

docs/reference/other-tools/sudo-rs.md
BAMF0: Ah yes, that is reasonable. These in-place definitions tend to be lost otherwise yes.

BAMF0: Fixed in: https://github.com/canonical/ubuntu-server-documentation/pull/775/commits/e74bf614e51c240226771d6f11f9435a545a4645

docs/reference/other-tools/sudo-rs.md
BAMF0: Terminal blocks look much better!

Fixed in: https://github.com/canonical/ubuntu-server-documentation/pull/775/commits/e74bf614e51c240226771d6f11f9435a545a4645

docs/reference/other-tools/sudo-rs.md
```diff
@@ -1,13 +1,130 @@
...
+The `rust-coreutils` package is available for installation on Ubuntu 24.04 LTS
+and later.
+
+Refer to the [uutils Coreutils Documentation](https://uutils.org/coreutils/docs/index.html) for information about its usage.
+
+### Release-specific exceptions
+
+The `rust-coreutils` package provides implementations of all `coreutils`
+utilities, which can also be invoked through its `coreutils` multi-call binary.
+However, starting with Ubuntu 25.10, the `coreutils-from-<provider>` packages
```
BAMF0, line 44: Don't like the "However" part here, it's unnecessary

s-makin: Can skip it :)

PR #788, opened 2026-08-31
typo in virtual-machine-manager.md

106106:
This line is no longer needed as the example in the CLI code block is now generic. (i.e. use <guestname>) and no longer mentions "web_devel".


s-makin (approved):
Well spotted, thanks!

PR #790, opened 2026-09-04
Minor improvements to Rails page

### LINKED ISSUE #560: [Feedback]: Improvements to Rails page
### Page URL

https://documentation.ubuntu.com/server/how-to/web-services/install-ruby-on-rails/

### Select an option

- [x] I found the information, but it was incorrect
- [x] I found the information, but it was incomplete

### Issue Description

Rails seems to be broken (I could nor run "rails new" nor install all the new apps deps with bundler). This should be re-visited. It would also be nice to include the OPTIONAL step to create a new app so a reader could actually run the commands presented on the page without having an existing app.

guilherme-puida:
### Description

Minor improvements to the Rails page. 

---

### Related issue

Fixes #560

---

### Checklist

- [x] I have disclosed my usage of AI

---

I feel like this page documents a really old-school method of developing/deploying Rails applications. I imagine most ruby developers don't use Rails from the archive (and probably not even the Ruby interpreter). I wonder if this is even useful in 2026, and maybe we should consider either completely revamping and modernizing this page, or just dropping it completely. In the meanwhile, this PR at least makes the commands work on a fresh Resolute system.

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -31,6 +31,7 @@ btrfs
 Bufferbloat
 bufferbloat
 bugfixes
+bundler
 Byobu
 Canonical's
 CAs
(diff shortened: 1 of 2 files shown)


athos-ribeiro (approved):
LGTM

s-makin (approved):
Nice, lgtm :)

PR #795, opened 2026-09-15
manage-software: refer to enabling proposed

cpaelzer:
We used to have such content, it is correct to be in the general project docs but we should make it findable and refer to it.

The documentation in the project docs is also the right recent variant which uses low pin priority (the default in later releases) to test from proposed more consciously.

Note: spotted while analyzing kernel docs around proposed enablement, which need to become a sprint breakout


s-makin (approved):
Fine by me, thanks :)

PR #796, opened 2026-09-17
Add note instructing users to unininstall NVIDIA DOCA before installing ours (or vice versa)

MitchellAugustin:
### Description

While our DOCA-OFED packaging is set up such that it should safely `conflict` with DOCA from NVIDIA's repository, we should still encourage users to explicitly uninstall any DOCA installed from an external repository first if their machine has it installed, since we can't validate that the external repository doesn't do something unexpected.

---

### Commit message for squash merge

Add note instructing users to unininstall NVIDIA DOCA before installing ours (or vice versa)

---

### AI usage

None

---

### Checklist

- [X ] I have read and followed the [Ubuntu Server contributing guide](https://ubuntu.com/server/docs/contributing/).
- [X ] My pull request is linked to an existing issue (if applicable).
- [X ] I have tested my changes, and they work as expected.
- [X ] I have disclosed my usage of AI

---

I think this is minor enough to fit the criteria of a "very small issue" and have not created an associated issue.

---


docs/how-to/networking/install-doca-ofed.md
```diff
@@ -38,6 +38,11 @@ The following DOCA-OFED metapackages are provided:
 
 ## Install DOCA-OFED
 
+:::{warning}
+To avoid possible conflicts, if you've already installed DOCA-OFED from NVIDIA's repository, please uninstall them before proceeding.
```
s-makin:
```suggestion
To avoid possible conflicts, if you've already installed DOCA-OFED from NVIDIA's repository, please uninstall it before proceeding.
```

s-makin (changes requested):
Thanks for adding this. I have one small nit.

In addition to the notice to uninstall, it would be good if we could provide the  command to check what you currently have installed + the uninstall step here - since the page itself doesn't provide it already.

MitchellAugustin:
> In addition to the notice to uninstall, it would be good if we could provide the command to check what you currently have installed + the uninstall step here - since the page itself doesn't provide it already.

@s-makin so, I could add what it would _probably_ be - but the issue here is that, since we don't actually have any control over the upstream DOCA repository, we do not necessarily know that the removal command we would state in this doc today would always be accurate. The closest option I can think of would be to say
`sudo apt remove doca-ofed` (which is what they call their top-level package today) and then `sudo apt autoremove` - but strictly speaking, `autoremove` may do more than just uninstall the doca packages, so I'm not sure if you would want that. (and the list of dependencies of `doca-ofed` is ever-changing, so the command to manually remove all of them won't always necessarily be consistent if we didn't use autoremove).

Let me know what you think. Thanks!

